PR #1, opened 2018-11-29
Login As Different User

andjsmit:
Adds ability for admins to switch to any user.


spec/features/log_in_as_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    User.new(user_attributes) { |u| u.save(validate: false) }
+  end
+
+  before do
+    login_as user
+  end
+
+  scenario 'Non-admin user is not allowed to see switch user form' do
+    visit '/users/sessions/log_in_as'
+    expect(page).to have_no_selector('select')
```
dlpierce:
```suggestion
    expect(page).to have_no_selector('select#switch_user_identifier')
```

PR #4, opened 2019-01-08
[ESSI-9] For holding location, use IUCAT API as authority

aploshay:
Should be good for a first iteration, but known things to revisit include:
- Adding new options for display elements, like url
- Modifying label display in editing select field vs display (specifically, keeping/dropping "Bloomington - " prefix)
- Confirming final form of the API once released

### DIFF
--- a/app/authorities/qa/authorities/iucat_libraries.rb
+++ b/app/authorities/qa/authorities/iucat_libraries.rb
@@ -1,0 +1,33 @@
+module Qa::Authorities
+  class IucatLibraries < Base
+    include WebServiceBase
+
+    def search(id)
+      all.select { |library| library[:code].match(id.to_s) }
+    end
+
+    def find(id)
+      data_for(id)[:library] || {}
+    end
+
+    def all
+      data_for('all')[:libraries] || []
+    end
+
+    private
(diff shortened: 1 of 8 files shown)


app/authorities/qa/authorities/iucat_libraries.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    def all
+      data_for('all')[:libraries] || []
+    end
+
+    private
+      def api_url
+        { development: 'http://iucat-feature-tadas.uits.iu.edu/api/library',
+          test: 'http://iucat-feature-tadas.uits.iu.edu/api/library',
+          production: 'http://iucat.iu.edu/api/library' }.with_indifferent_access[Rails.env]
```
randalldfloyd: Would it be too late before sprint end on 1-15 to parametrize the values for api_url via config/secrets?

aploshay: @randalldfloyd Not at all -- I should have done that step before putting up the PR

aploshay: @randalldfloyd added a commit

PR #6, opened 2019-01-17
ESSI-30 add IU-branded header and footer in the application

wang58:
Added IU branded header and footer. They are similar as pages.dlib.indiana.edu.

### DIFF
--- a/app/views/shared/_footer.html.erb
+++ b/app/views/shared/_footer.html.erb
@@ -1,15 +1,14 @@
-<footer class="navbar navbar-inverse site-footer">
-  <div class="container-fluid">
-    <div class="navbar-text text-left">
-        <p><%= t('hyrax.footer.service_html') %></p>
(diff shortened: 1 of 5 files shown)


app/views/shared/_footer.html.erb
```diff
@@ -1,15 +1,14 @@
-<footer class="navbar navbar-inverse site-footer">
-  <div class="container-fluid">
-    <div class="navbar-text text-left">
-        <p><%= t('hyrax.footer.service_html') %></p>
-        <p><%= t('hyrax.product_name') %> <%= link_to "v#{Hyrax::VERSION}", "https://github.com/samvera/hyrax/tree/v#{Hyrax::VERSION}" %>
-          ESSI <%= link_to(ENV['SOURCE_COMMIT'], "https://github.com/IU-Libraries-Joint-Development/essi/tree/#{ENV['SOURCE_COMMIT']}") if ENV['SOURCE_COMMIT'] %></p>
```
dlpierce: I would like the display of the current commit to be retained in the footer.

wang58: Thanks! Good idea, I will it in.

PR #11, opened 2019-02-12
Adds @ to default email address when not fetched from ADS.

andjsmit:


wang58:
The current default_email_domain sets to " indiana.edu" which is in "essi_config.example.yml" line 35, I suggest to change to "iu.edu" according to the description of the Jira.

wang58:
The changes look good and straightforward. I will merge the pull request.

PR #14, opened 2019-02-15
ESSI-74 Fixes inability to add a PagedResource in the staging instance

randalldfloyd:
In the Docker staging environment, when you go to add a PagedResource you immediately get a generic error as soon as the edit form starts to render.  This is due the secrets configuration file not having a value for the url of the holding location authority service, resulting in a method being called on a nil result.

The configuration obviously needs to have that value added, but this PR adds a guard clause in the authority class to just return an empty object if the service url doesn't exist.  The resulting holding locations dropdown will be empty but at least the edit form can render.

### DIFF
--- a/spec/authorities/qa/authorities/iucat_libraries_spec.rb
+++ b/spec/authorities/qa/authorities/iucat_libraries_spec.rb
@@ -5,6 +5,17 @@ RSpec.describe Qa::Authorities::IucatLibraries do
   let(:matching_id) { 'B-WELLS' }
   let(:nonmatching_id) { 'foobar' }
 
+  context 'when configuration does not exist' do
+    describe '#all' do
+      let(:result) { authority.all }
+      it 'returns an empty Array' do
+        allow(ESSI.config).to receive(:[]).with(:iucat_libraries).and_return(nil)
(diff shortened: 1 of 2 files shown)


spec/authorities/qa/authorities/iucat_libraries_spec.rb
```diff
@@ -5,6 +5,19 @@
   let(:matching_id) { 'B-WELLS' }
   let(:nonmatching_id) { 'foobar' }
 
+  context 'when configuration does not exist' do
+    describe '#all' do
+      let(:result) { authority.all }
+      it 'returns an empty Array' do
+        iucat_libraries_url = ESSI.config[:iucat_libraries]
```
aploshay: An alternative to the 3 lines for stashing, nullifying, and restoring the config value is:
`allow(ESSI.config).to receive(:[]).with(:iucat_libraries).and_return(nil)`
but I'm not sure which approach is more Hyrax-y

randalldfloyd: @aploshay I pushed a new commit with the much improved test.

PR #16, opened 2019-02-21
[ESSI-23] Create hOCR with image derivatives

aploshay:
This addresses ESSI-23.  Running this commit, after you upload an image file as a FileSet, that image should be available as the `original_file`, and the resultant hOCR as `extracted_text`.  (You will need `Tesseract` installed, with support for English language.)

Broadly, what this does is add hOCR generation following the current Hyrax pattern for this sort of thing:
- A `Runner`, whose job is to specify the `Processor`, (optionally) services for input and output files, and has an option to modify the incoming directives before they are passed on.  You can see [some stock examples and the parent class in hydra-derivatives](https://github.com/samvera/hydra-derivatives/tree/master/lib/hydra/derivatives/runners).
- A `Processor`, whose job is to actually create the derivative(s).

This also includes some related infrastructure:
- a `Tesseract` class for handling language values
- a customization of `PersistDirectlyContainedOutputFileService`, because I ran into some breaking problems using the [stock Hyrax version](https://github.com/samvera/hyrax/blob/master/app/services/hyrax/persist_directly_contained_output_file_service.rb) (based off of [PersistBasicContainedOutputFileService](https://github.com/samvera/hydra-derivatives/blob/master/lib/hydra/derivatives/services/persist_basic_contained_output_file_service.rb) in Hydra::Derivatives) -- detailed further below
- an initializer to modify the stock `FileSetDerivativesService`, to add a call to make hocr
- an initializer to modify _all_ derivative creation to change the default `source_file_service` to a stock remote retrieval one (instead of the stock local retrieval one), if we're storing original files in Fedora.

Some notable issues I ran into:
- The [`Hydra::Derivatives` logic for determining `mime_type` and `original_name`](https://github.com/samvera/hydra-derivatives/blob/master/lib/hydra/derivatives/services/persist_output_file_service.rb#L12-L28) was heavily broken for us, because it just checks that the source file _responds_ to `original_filename` (which actually also generates deprecation warnings -- it's to be replaced by `original_name`) and `mime_type` methods, and then uses those results if they're available.  The breaking problem was that, in our case, although the incoming `IoDecorator` had those methods, one was providing a blank string, the other a nil -- and the resulting `Hydra::PCDM::File` would then save fine, but break the parent `FileSet` saving, and _also_ prevent its deletion.
We're working around this with our customized output file service, but we should probably send this as an issue, up the stack?

- Attaching an .hocr file without specifying a UTF-8 character encoding was assuming an ASCII 8-BIT encoding -- even when it contained UTF-8 characters outside the ASCII 8-BIT set -- which was then breaking `FileSet` save, again.
We're working around this by specifying a UTF-8 character set, but it might also be an issue worth addressing higher up in the stack?

- Locally, I'm getting intermittent issues running the tests when it calls `FactoryBot` to generate a `FileSet`, _sometimes_ getting the ldp error "can't call create on existing resource".  But I think this is a known current issue?

- This does not include a spec for the customized `PersistDirectlyContainedOutputFileService`, because I'm running into a weird issue where the code works fine in development but the tests produce an `ActiveFedora` error.  I didn't want to to hold back the story, so I left it out.  But I could add it back in as skipped tests, and/or someone can look into this more and see how to try to fix whatever the root issue turns out to be.


andjsmit:
+1

Was able to ingest newspaper page. hOCR file was created and stored in Fedora.
Tested multiple languages against Tesseract service. Got expected results.
Ran spec tests twice. Failed once as described above, second time all test passed.

PR #22, opened 2019-03-15
ESSI-121 Adds OCR indicators in file manger and download link on file details.

andjsmit:


rwingiupui:
👍 +1

config/initializers/add_extracted_text_downloads.rb
```diff
@@ -0,0 +1,29 @@
...
+Rails.application.config.after_initialize do
+  Hyrax::DownloadsController.class_eval do
+    prepend DownloadsControllerExtensions
+  end
+end
+
+# Makes extracted text check available in file details.
+module FileSetPresenterExtensions
+  def extracted_text?
+    return true if ::FileSet.find(id).extracted_text
```
aploshay: could condense to 1 line with: `::FileSet.find(id).extracted_text.present?` ?

app/views/hyrax/file_sets/_show_details.html.erb
```diff
@@ -0,0 +1,23 @@
...
+<dl class="dl-horizontal file-show-term file-show-details">
+  <dt>Depositor</dt>
+  <dd itemprop="accountablePerson" itemscope itemtype="http://schema.org/Person"><span itemprop="name"><%= link_to_profile @presenter.depositor %></span></dd>
+  <dt>Date Uploaded</dt>
+  <dd itemprop="datePublished"><%= @presenter.date_uploaded %></dd>
+  <dt>Date Modified</dt>
+  <dd itemprop="dateModified"><%= @presenter.date_modified %></dd>
+  <% if @presenter.extracted_text? %>
+  <dt>OCR</dt>
+  <dd><%= link_to 'Download', "/downloads/#{@presenter.id}?file=extracted_text" %></dd>
```
aploshay: is there a more orthodox way to build this URL?  That's a genuine question; I don't object to what's here currently

config/initializers/add_extracted_text_downloads.rb
andjsmit: Yes - reads better - thanks.

app/views/hyrax/file_sets/_show_details.html.erb
andjsmit: Not sure of a better way to form URL, but I moved the link creation to the presenter so all extracted_text stuff resides in same file.

PR #23, opened 2019-03-19
Essi 34 facet indexing

dlpierce:
Add publisher, publication place, date created pages and state facets.
Enables PagedResource feature spec


dlpierce:
This should be ready for merge now.

PR #24, opened 2019-03-21
[ESSI-171] Make source_metadata_identifier optional

aploshay:
This resolves ESSI-171 by making all of `title`, `creator`, and `source_metadata_identifier` optional in the form submission.  If all are left blank, form submission will fail and return a user validation error that Title is required (since it was neither provided directly, nor returned by a remote metadata lookup).  This should be sufficient to resolve ESSI-171 and we can potentially revisit later, if desired, having better UI behavior with regards to indicating functional mutex sets for creator + title vs source_metadata_identifier.

I should note this smuggles in more monkeypatched refactoring than you might expect for a bugfix, as it modifies base `Hyrax::Forms::WorkForm` behavior to make `primary_terms` explicitly configurabe in the form model, instead of assumed to be equivalent to primary fields.  This basically follows the existing pattern for specifying `terms` and `required_fields` -- with the difference that the base class starts with an empty Array of values for `primary_fields`, instead of a default set.  (I can change this to be more consistent, and start with an assumed set of values equivalent to the standard required fields; I just preferred it this way, as it feels cleaner to me, but I guess I'm coming from a different mindset.)


aploshay:
Note that the failing test is for an unrelated spec, with:
```
 Ldp::Conflict:
       Can't call create on an existing resource
```
which is a known, separate, issue?

andjsmit:
Ran into an issue when saving without title: 

(1) Reloaded page with notice of title requirement
(2) Files were no longer included
(3) Upon adding files again and saving, error occurred:

Unknown agent type ""

from line: https://github.com/IU-Libraries-Joint-Development/essi/blob/master/app/actors/essi/actors/create_with_remote_files_actor.rb#L17

aploshay:
@andjsmit weird, I'm unable to recreate this error following your instructions.  Do you get different behavior doing the same steps on the master branch?

randalldfloyd:
@andjsmit or @aploshay Was the discussion about this PR resolved?  Should this be merged?

PR #26, opened 2019-03-26
[ESSI-178] Update gemfile to hyrax 2.5.0

andjsmit:


Gemfile
```diff
@@ -69,7 +69,7 @@ end
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 
-gem 'hyrax'
+gem 'hyrax', '2.5.0'
```
aploshay: Do we want to lock to 2.5.0, specifically, and not 2.5.x ?

andjsmit: It is my thinking that it would be safer to lock to the specific version so updates do not take us by surprise. The past minor release (2.4.x) only made it to 2.4.1, so I don't think we need to worry about constantly updating. But I'm open to changing if deemed best.

aploshay: Updates shouldn't take us by surprise, since we still have to run a `bundle update` to pick up a 2.5.1 or upwards, so I vote for `~> 2.5.0`.

andjsmit: OK. Will change.

PR #27, opened 2019-03-29
Bundle update

aploshay:


andjsmit:
+1 - Application worked as expected after fetching branch and running bundle update.

PR #30, opened 2019-04-02
[ESSI-36] Apply some #save_structure bugfixes to ESSI (as was to pumpkin)

aploshay:
Note that completion of ESSI-24 is also required for full resolution of `#save_structure` bugs, pertaining to job queuing.

This covers:
* aborting a `#save_structure` request if the resource is locked, returning 423 status.
* sharing a `NodeCache` across nested `OrderedList` members, to avoid id collision


aploshay:
I pushed a change that resolved the failing tests, except for the one that gets a 423 back when calling `#save_structure` on a locked resource.  But I think the issue may actually be with the lockable behavior, itself, as I can't get it to work in console... anyone else?

PR #31, opened 2019-04-30
[ESSI-139] In the File Manager, add Viewing Direction and Viewing Hint

aploshay:
Required adding and modifying a couple of stock hyrax assets: one javascript, one view partial.

Since a model's form `terms` are used for _both_ deciding what passes `params` sanitizing _and_ what shows up in the general editing form, this does add `viewing_direction` and `viewing_hint` to the Edit form, as a side effect.  In the longer term, we probably want to either hide/remove it, or change the interface from free text to a select box.

One thing this revealed is that any additions you want to make to File Manager sidebar content will necessarily require modifying the stock Hyrax javascript, if you want to the "Save" button-enabling logic to be aware of your new content -- otherwise, users will have to change something _else_ to get the Save button enabled!


aploshay:
@randalldfloyd Just force-pushed a rebase onto a fresh master, and slightly tightened up the commit history; no changes to the final result (at least, none intended)

PR #35, opened 2019-05-14
[ESSI-176] Search extracted text within UV

aploshay:
Adds support for searching extracted text within UV, and getting annotation layer highlighting of each result.

The last set of commits -- adding derivation of word boundaries, and parsing it for use to determine the bounds for annotation highlighting -- work as-is.  But merits future review/refactoring to be in keeping with what's going on in Newspaper Works and ALTO, and for performance improvement.  Word boundary creation in derivatives also merits test coverage, but I figure that can await the review/refactoring?


aploshay:
Regarding commit 7f46396, "Add workaround for webmock/vcr body issue (HPT-1561)":

In JIRA, I've added this story as HPT-1561:
Stack bug: output_file_service fails tests (only) when Hydra::Derivatives::IoDecorator is given a raw String, instead of a File.open call

In hydra-derivatives, the spec for the full_text processor implies we should be able to call an output file service built on a raw String:
https://github.com/samvera/hydra-derivatives/blob/master/spec/processors/full_text_spec.rb

In practice, doing so works in development but breaks the test stack with webmock/vcr, where at some point a Hydra::Derivatives::IoDecorator value instead of a raw String is passed as a body, and fails this check:
https://github.com/vcr/vcr/blob/master/lib/vcr/structs.rb#L61-L74

Ironically, building the Hydra::Derivatives::IoDecorator on a File.open call eventually results in the webmock/vcr request having a raw String value for the body.  So somewhere in between, there's a String conversion happening that's not robust for the case of building Hydra::Derivatives::IoDecorator on a String, in the first place.

PR #44, opened 2019-06-06
essi-137 bulk page view action

wang58:
The bulk action will show when one or several images/resources are picked.


andjsmit:
+1 : No Test failures for me locally and the bulk actions block worked as expected.

rwingiupui:
+1 : The bulk actions block feature worked fine and all local tests completed without any failures.

PR #46, opened 2019-06-21
[ESSI-253] Display metadata in UV right sidebar.

andjsmit:
Fixes IIIF manifest issue with not containing metadata if missing afield value. Also expands fields to match display attributes.


wang58:
The functionality is good. I saw the metadata displayed in the UV right sidebar correctly.

wang58:
The code looks good to me too. I will merge the code to master.

PR #47, opened 2019-06-24
ESSI-279: Fixing problem suddenly breaking all commits and PRs

randalldfloyd:
Updates selenium-webdriver and capybara on the client side to fix incompatibility with headless chrome on CircleCI.  

At some point recently, commits in the repo started failing, even for previously passing PRs when master was merged back in.  This was ultimately determined to be caused by a sudden incompatibility between the server-side headless chrome and the client-side version of the Selenium driver. This was causing failures in the Paged Resource feature test when Javascript interactions were initiated.  

The underlying root cause was incompatible response formats between the client and server when checking for w3c standards compliance.  Updating the Selenium webdriver fixes this, but this then produces many Selenium deprecation errors which are fixed by updating Capybara.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -41,7 +41,7 @@ group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
   # Adds support for Capybara system testing and selenium driver
-  gem 'capybara', '~> 2.13'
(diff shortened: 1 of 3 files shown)


randalldfloyd:
See also: 
https://stackoverflow.com/questions/56111529/cannot-call-non-w3c-standard-command-while-in-w3c-mode-seleniumwebdrivererr

PR #48, opened 2019-06-25
Resolve possible null value for default_url_options to empty hash

aploshay:
If this assignment returns nil, things get broken, so this adds a nil guard to resume the normal default value of an empty hash.

### DIFF
--- a/config/environment.rb
+++ b/config/environment.rb
@@ -5,4 +5,5 @@ require_relative 'application'
 Rails.application.initialize!
 
 
-ESSI::Application.default_url_options = ESSI::Application.config.action_mailer.default_url_options
+
+ESSI::Application.default_url_options = ESSI::Application.config.action_mailer.default_url_options || {}
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -30,6 +30,7 @@ Rails.application.configure do
   config.action_mailer.raise_delivery_errors = false
 
   config.action_mailer.perform_caching = false
+  config.action_mailer.default_url_options = { host: "localhost:3000" }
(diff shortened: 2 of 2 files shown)


aploshay:
@randalldfloyd when [this change was initially made](https://github.com/IU-Libraries-Joint-Development/essi/commit/80b2af7ca7373fec2d722f7d87f01a464805b872#diff-1d98f0039bbc6df92adee882ed99d993L68), do you know why it got values for the production and test environments, but not development?
[relevant commit](https://github.com/IU-Libraries-Joint-Development/essi/commit/80b2af7ca7373fec2d722f7d87f01a464805b872#diff-1d98f0039bbc6df92adee882ed99d993L68)

andjsmit:
I'm finding that this change may break the iiif manifest creation in development as `default_url_options[:host]` is not set.

aploshay:
> I'm finding that this change may break the iiif manifest creation in development as `default_url_options[:host]` is not set.

@andjsmit Can you expand on this?  What conditions do we need to meet to resolve the original issue without creating the one you cite here?

randalldfloyd:
@andjsmit Was there something more we need to do to consider merging this PR?

andjsmit:
@aploshay  @randalldfloyd  What I was running into is that, as is, setting `default_url_options` to an empty hash caused the UV not to load for me because `default_url_options[:host]` was missing. Apparently, this missing value was needed in the iiif manifest creation. This appears to be a development environment only issue and easily fixed by setting the value in configuration. So if that works for everyone, I'm not opposed to merging as is.

PR #51, opened 2019-07-11
ESSI-286 Make universal Viewer size taller

wang58:
According to the hyrax style, I put 10px padding in. Now the uv size plus padding is 800px.

### DIFF
--- a/app/assets/stylesheets/hyrax/_viewer.scss
+++ b/app/assets/stylesheets/hyrax/_viewer.scss
@@ -1,0 +1,4 @@
+.viewer_iiif {
+  height: 800px;
+  padding: 10px;
+}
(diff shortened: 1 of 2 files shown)


aploshay:
When we override stock components from hyrax gems, I really like the pattern of having one commit for just adding the original file directly into the local repo, and a separate commit for the changes we make, to allow for for easier diff'ing and tracking how and why we made changes.  I'd like to amend this PR by rebasing it and squeezing in a separate commit to pull in the [universal viewer partial](https://github.com/samvera/hyrax/blob/master/app/views/hyrax/base/iiif_viewers/_universal_viewer.html.erb); any +1's or objections?  (I know Ying is out right now, but I also don't want this stalled out in the meantime.)

PR #53, opened 2019-07-15
Admins are always in registered user group, in user#groups method

aploshay:
For ESSI-344, adds the corresponding change to the Ability model into the User model.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,9 +72,9 @@ jobs:
     # Download and cache dependencies
     - restore_cache:
         keys:
-        - dependencies-v1-{{ .Branch }}-{{ checksum "Gemfile.lock" }}
-        - dependencies-v1-{{ .Branch }}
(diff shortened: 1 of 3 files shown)


aploshay:
If having this logic in both models ([Ability](https://github.com/IU-Libraries-Joint-Development/essi/blob/master/app/models/ability.rb#L27-L34), [User](https://github.com/IU-Libraries-Joint-Development/essi/blob/master/app/models/user.rb#L39-L43)) is redundant, we could potentially remove one -- but they are distinct, though, and come from different parts of the stack ([Blacklight](https://github.com/projectblacklight/blacklight-access_controls/blob/master/lib/blacklight/access_controls/ability.rb#L99-L106), [hydra-role-management](https://github.com/samvera/hydra-role-management/blob/master/app/models/concerns/hydra/role_management/user_roles.rb#L12-L16)), and I figure as long as it's in both places they should be consistent

PR #55, opened 2019-07-18
ESSI-288: editor layout consistency

aploshay:
Modifies button layout in Show page, and Edit, Edit Structure, and File Manager views for consistency.

A couple of included changes worth noting:
* Modified the CSS for the fixedsticky class with greater specificity, since it was actually broken out of the box.  (Now, it actually _is_ displayed as fixed.)
* Modified breadcrumb support to also work for `#file_manager` and `#structure` actions
  * This obviated the need for `structure_helper.rb`, now removed


aploshay:
Update on broken specs: I believe the ul, li count changed because we switched from the custom breadcrumbs which was <ul> to the standard one, which is <ol>.  So I just adjusted the expected count to match the new reality.

aploshay:
[Hyrax issue 3891](https://github.com/samvera/hyrax/issues/3891) raised for broken `fixedsticky` CSS class

PR #62, opened 2019-08-08
[ESSI-360] Provide configuration option to disable derivative creation.

andjsmit:


dlpierce:
We seem to be collecting a large amount of code living in initializers, which should be a last resort.

Would it be possible to move the code that is being prepended into the FileSetDerivativesService into its own derivative service? Then that service class could be appended onto `Hyrax::DerivativeServices.services`?

spec/services/hyrax/file_set_derivatives_service_spec.rb
```diff
@@ -1,9 +1,9 @@
 require 'rails_helper'
 
-RSpec.describe Hyrax::FileSetDerivativesService do
+RSpec.describe Hyrax::DerivativeService do
```
andjsmit: @dlpierce Should this file then get renamed `DerivativeService`?

dlpierce: Yeah, I'll do that.

PR #63, opened 2019-08-15
ESSI-385 Allows Sidekiq to be configurable via environment settings in secrets file

randalldfloyd:
Allows Sidekiq queue names and failed job retry count to be configurable via main ESSI config file.

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -1,4 +1,2 @@
 ---
-:queues:
-  - default
-  - ingest
+:queues: <%= ESSI.config.fetch(:sidekiq,{}).fetch(:queue_namess,nil) || "['default','ingest']" %>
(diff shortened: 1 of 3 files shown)


config/sidekiq.yml
```diff
@@ -1,4 +1,2 @@
 ---
-:queues:
-  - default
-  - ingest
+:queues: <%= ESSI.config.fetch(:sidekiq,{}).fetch(:queue_names,nil) || "\n- default\n- ingest\n" %>
```
dlpierce: An inline array might be clearer.
```suggestion
:queues: <%= ESSI.config.fetch(:sidekiq,{}).fetch(:queue_names,nil) || "['default', 'ingest']" %>
```

randalldfloyd: @dlpierce Requested changes made and pushed.

PR #64, opened 2019-08-16
Redefining how and where Sidekiq configuration values get set

randalldfloyd:
PR #63 introduced the ability to configure Sidekiq settings from the main ESSI configuration YAML file.  However, it wasn't until trying to run Sidekiq standalone that it was apparent that the ESSI config stash was not loaded and available when Sidekiq's default config file is parsed (`config/sidekiq.yml`)  which resulted in a runtime error (i.e. uninitialized constant ESSI)

Rather than try to figure out how to provide the ESSI config stash to that file, the Sidekiq initializer was refactored to have complete control of the assignment process and set an order of precedence.  See comments in `config/initializers/sidekiq.rb`.


randalldfloyd:
@dlpierce A refactoring of how Sidekiq options are set ready for review

PR #65, opened 2019-08-20
Docker based dev env

orangewolf:
This is pretty close to working. I was able to get all the services up, and I found the special solr config required for blacklight-iiif. However, when adding an item to Fedora it does not seem to automatically create the /dev base path (same for test). On other Hyrax apps I see
```
ActiveFedora: loading fedora config from /data/config/fedora.yml
ActiveFedora: loading solr config from /data/config/solr.yml
Attempted to init base path `dev`, but it already exists
```

On this app I see
```
ActiveFedora: loading fedora config from /data/config/fedora.yml
ActiveFedora: loading solr config from /data/config/solr.yml
```

And then the item gets an Ldp::NotFound issue for /dev.

If anyone has ideas on what might cause this, I'm currently trying to figure it out.


orangewolf:
The above is resolved. Everything is working now except for the Javascript specs

PR #66, opened 2019-08-20
[ESSI-285] Import manifest caching from pumpkin

aploshay:


aploshay:
Note that to activate this in development, this file needs to exist in your project tree:
`tmp/caching-dev.txt`
to satisfy [this condition check](https://github.com/IU-Libraries-Joint-Development/essi/blob/master/config/environments/development.rb#L15-L27).

app/models/resource_identifier.rb
```diff
@@ -0,0 +1,33 @@
...
+class ResourceIdentifier
+  attr_reader :id, :repository
+  def initialize(id, repository = nil)
+    @id = id
+    @repository = repository || SolrRepository
+  end
+
+  # Unique key of the object pointed to by ID's state.
+  def to_s
+    @digest ||= Digest::MD5.hexdigest(timestamp)
```
dlpierce: Seems like a unique key should take more into account than just a timestamp. Any reason to not also include the id in the hash?
```suggestion
    @digest ||= Digest::MD5.hexdigest(@id + timestamp)
```

PR #69, opened 2019-08-21
Try to fix flaky feature spec on add new work selection modal

dlpierce:


dlpierce:
This has passed 4 runs on Circle with no fails.

PR #75, opened 2019-09-03
Essi 119 full screen

wang58:
The huge part of code is copied over the whole uv-2.0.1 library from the pumpkin. The structure of this functionality is similar to the pumpkin.


wang58:
@geekscruff, I still did not figure out how to port the latest hyrax code.  Could you give more details, like which part I need add or remove from my current files?

geekscruff:
Sorry for not replying sooner, have a look at this PR (ignore the fits references):
https://github.com/antleaf/nims-hyrax/commit/d4d65ed8d6851e4b355b165d3f9f55669fdd05a4

This would replace the local universal viewer with the yarn installed one, which would be deployed at build time. It would save on having all the code locally, and would also prepare ESSI for changes coming in Hyrax 3.

The steps would be to add / change:
* install yarn in the Dockerfile
* add config/uv/uv-config.json and 
* add app/helpers/hyrax/iiif_helper.rb and config/uv/uv.html (uv-config.json can be customized to enable/disable uv features)
* add app/views/hyrax/base/_representative_media.html.erb (or merge if you have a local copy of that file)
* add app/views/hyrax/base/iiif_viewers/_universal_viewer.html.erb
* add the new css classes to app/assets/stylesheets/hyrax/_viewer.scss  (and review what's already there - test this to get the style right)
* review other code to bring it in line to the new approach

Then you would be able to delete public/bower_includes/universalviewer

aploshay:
Note that this includes upgrading from `universal_viewer` v2 to v3.

dlpierce:
I think `yarn install` does need to be included in the Dockerfile near where assets are precompiled, and also in the dev image?

PR #76, opened 2019-09-06
Fixes enable/disable Save button in file manager.... 

andjsmit:
... by syncing text in JS and view with i18n.


andjsmit:
Moved to a new PR in order to split commits.

PR #78, opened 2019-09-11
ESSI-416 Adding ability to disable search within for a PagedResource

randalldfloyd:
A PagedResource will typically have searchable text extracted during OCR derivative creation, but there will be a chance that the fidelity of the text is not high enough to make searching useful.  In these cases we will want to hide the search bar in the Universal Viewer.  

This PR adds all of the pieces needed to disable searching via a control in the File Manager toolbar.  It also adds a step in the actor to ensure that searching is enabled by default so that the option only needs to be changed in the case of wanting to disable it.  All of this is specific to the PagedResource work type.

### DIFF
--- a/app/actors/hyrax/actors/paged_resource_actor.rb
+++ b/app/actors/hyrax/actors/paged_resource_actor.rb
@@ -6,6 +6,7 @@ module Hyrax
       def create(env)
         apply_creation_data_to_curation_concern(env)
         apply_save_data_to_curation_concern(env)
+        apply_ocr_option_data_to_curation_concern(env)
         save(env) && next_actor.create(env) && run_callbacks(:after_create_concern, env)
       end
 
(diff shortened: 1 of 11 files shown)


app/actors/hyrax/actors/paged_resource_actor.rb
```diff
@@ -3,6 +3,19 @@
 module Hyrax
   module Actors
     class PagedResourceActor < Hyrax::Actors::BaseActor
+      def create(env)
```
aploshay: You could break this up into two commits where you (1) copy in the inherited `#create` method from `Hyrax::Actors::BaseActor`, and then (2) add in the new line for `apply_ocr_option...`, if you want to follow that pattern for making changes to stock Hyrax behavior more cleanly isolated.

aploshay:
+1 by inspection.  Added a comment about potentially breaking one change into two commits.  (It needs a commit in any case, since I got bonked heads here and got the PR out-of-sync with master by making a change to `paged_resource_spec.rb` in mine.)

app/actors/hyrax/actors/paged_resource_actor.rb
randalldfloyd: Good idea, I split commits as suggested.  I tried to `super` the method to avoid that, but there was just something about the atomic nature of the save when broken up that just wouldn't work so I was forced to do the steps explicitly and in order before save.

PR #80, opened 2019-09-11
WIP ESSI-187, UI display of locked works

aploshay:
Marked WIP pending PR review.
* Adds display of lock warning into edit, structure, file manager views
* changes generation of `LockManager` to a single instance for standard use, and a single instance for `lock?` checks
   * gets speedier results of `lock?` checks

Doesn't:
* doesn't add a `lock?` check into the `save_structure` action as it was already there
* doesn't modify `LockableJob`

### DIFF
--- a/app/helpers/lock_warning_helper.rb
+++ b/app/helpers/lock_warning_helper.rb
@@ -1,5 +1,5 @@
 module LockWarningHelper
-  def lock_warning
+  def lock_warning(curation_concern)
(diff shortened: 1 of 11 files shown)


aploshay:
@dlpierce it should show up in feature specs, but we only have them right now, for create actions, not for edit, structure, or file manager where it would show up.  I have added other spec coverage.

PR #81, opened 2019-09-16
ESSI-457 Remove IUCAT Holding Locations from ESSI config

randalldfloyd:
The ESSI config contained the full configuration value for the IUCAT holding location, which usually works even in the developer environment even though that should probably be a deployment concern.  However, when outside developers began working with the app (i.e. Notch8) the service could not be contacted and created timeout errors.  This PR disables the api configuration by default and makes the QA vocab (and specs) behave correctly when not configured.


aploshay:
+1 by inspection

PR #82, opened 2019-09-18
[ESSI-412] Toggle Show / Hide for Child Items button.

andjsmit:


app/assets/javascripts/work_items.es6
```diff
@@ -0,0 +1,18 @@
+export default class WorkItems {
+  constructor() {
+    this.toggle_show_items()
+  }
+
+  toggle_show_items(){
+    let button = $("#work-items-button")
+    let show = I18n.t('hyrax.works.form.show_child_items')
```
randalldfloyd: Should there also be a translation by default in `en.hyrax` for this?  When I desk check I get `missing "en.hyrax.works.form.hide_child_items" translation`

andjsmit: Ooops. Missed committing locale file.

randalldfloyd: Works locally now.

randalldfloyd:
Merging with CI failure.  Changes work locally and the failure is the pesky LDP conflict.

PR #84, opened 2019-09-24
ESS-509 Forcing ActiveFedora::Cleaner on all rspec examples to avoid LDP Conf…

randalldfloyd:
In the past weeks, the test suite started failing over 70% of the time due to LDP Conflict errors. We used to get them occasionally but we ignored them and restarted builds.  But something got added into the test suite that further amplified some underlying problem in our tests.  For now we can eliminate the pain point by forcing `ActiveFedora::Cleaner` before all rspec examples, but we need to figure out what the pattern in a Hyrax/ActiveFedora app should be or what test(s) is triggering the problem.


dlpierce:
This PR now sets the noid minter to test mode while running specs, instead of forcing a clean before every example.

PR #86, opened 2019-09-27
Essi 413 uv download

wang58:
The change will fix both ESSI-413 (UV download current view image) and ESSI-414 (UV download original image) issue.

### DIFF
--- a/.solr_wrapper
+++ b/.solr_wrapper
@@ -1,5 +1,5 @@
 # Place any default configuration for solr_wrapper here
-version: 7.7.1
+# version: 7.7.1
 # port: 8983
 instance_dir: tmp/solr-development
 collection:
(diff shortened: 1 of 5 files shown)


wang58:
Ok, See the extra commits for the override. 
44bb1b6... is the override file copied from hyrax
66970cd... is the changes.

aploshay:
In ee17ee5, I pulled in the Hyrax 3.x performance improvements based on solrizing `original_file_id`, but not those pertinent to the alpha channels, and worked around updating `VersioningService`.

wang58:
Updated the branch, the uv download worked as expected. The code looked good to me. +1

aploshay:
@randalldfloyd Can you close or update your change request?  Thanks.

PR #90, opened 2019-10-02
ESSI-84 When navigating from search results to show page, invoke UV search with term

randalldfloyd:
Replicates functionality in Pumpkin that automatically passes search term from a results link into the UV and invokes search.  

There is new JavaScript in the Universal Viewer partial that looks for the `query` parameter and uses its value, if it exists, to invoke the UV search within.  It has logic to determine when the UV search box elements are in place before trying to act upon its selectors, which can take a variable amount of time depending on manifest building and/or caching.


randalldfloyd:
> There's a TODO note -- do we want a corresponding story in the backlog?

That may be an artifact from overriding that piece from upstream Hyrax.  I picked this up in-progress so I didn't get a chance to commit the original file first before modifying it.

PR #93, opened 2019-10-07
Enforce font-size override in branding CSS

aploshay:
Enforces the override introduced in fb6e88cfa


wang58:
The font looks good to me. Approve this change.

PR #96, opened 2019-10-24
Essi 549 side options menu

wang58:


app/assets/javascripts/hyrax/file_manager.es6
```diff
@@ -5,7 +5,7 @@ export default class FileManager {
   constructor() {
     this.save_manager = this.initialize_save_manager()
     this.sorting()
-    this.save_affix()
+//    this.save_affix()
```
aploshay: @wang58 Why is this line commented out, and retained?

wang58: Removed the line.

aploshay:
@wang58 Why did you remove the `save_affix` call?  Or, why was it there, originally?

wang58:
the method save_affix dynamically added some css in. On the top, it is “affix-top” and in the end, it is “affix”. This caused the width differently.  I am not sure why we need “save_affix”, it is not from hyrax or pmp. Removing it will make the width consistent and that is the jira for.

wang58:
file_manager.es6 and save_affix() is from ESSI-139. @aploshay, you added it in and knew more about save_affix() for. Please remove save_affix() cause some regression, please let me know.

aploshay:
@wang58 Some more on this:
The save_affix() behavior is inherited from stock hyrax.  So, I _think_ the preferred approach here is to fix the css for the affix-top/affix/affix-bottom classes, instead of cutting out that behavior.  But we also do see the width-changing behavior in our vanilla hyrax instance:
https://demo.ulib.iupui.edu/concern/images/xk81jk370/file_manager?locale=en
so if we fix this in ESSI we might want to submit the same change in Hyrax.

wang58:
@aploshay That's true the save_affix() behavior is inherited from hyrax. I agree another approach might adjust the css. I did not see affix-top, affix-bottom classes in any place, so I guess they might be placeholders? I do found .affix class in 
 /* line 128, /Users/wang58/.rvm/gems/ruby-2.6.2/gems/hyrax-2.5.0/app/assets/stylesheets/hyrax/_file_manager.scss */
  *[data-action=file-manager] #file-manager-tools.affix {
    width: 285px;
  }
I want the browser ignore this width but did not find a good way to override it. If you have a good way for the css, let me know.

wang58:
Ok, find another way to have the width consistent by changing css only.

wang58:
ok, the code got reviewed and passed all checks. I will merge it to master.

PR #97, opened 2019-10-30
[ESSI-516] Fix broken javascript on essi-stage

aploshay:
Including `jasmine-jquery.js`, needed for javascript spec purposes, in the main assets tree was breaking javascript -- but only when assets were compiled and minified.  (So, everything was running fine "on my machine" in development mode, but not on essi-stage.)

This PR moves that file to the spec tree, where it's still picked up for inclusion in tests, and drops the explicit include.  This gets us consistent with how Hyrax handles this:
https://github.com/samvera/hyrax/blob/master/spec/javascripts/helpers/jasmine-jquery.js
https://github.com/samvera/hyrax/blob/master/spec/javascripts/support/jasmine.yml#L40-L41


aploshay:
@bkiahstroud tagging you for the conclusion to our slack conversation about this

dlpierce:
@geekscruff  No, we wanted to let you know that the previous method of loading `jasmin-jquery.js` needed some adjustment in case it was also used on other projects.

bkiahstroud:
@aploshay Ahh, good to know. I didn't realize Jasmine was in Hyrax, I didn't even think to look there for inspiration. Thanks for bringing this full circle, I was curious about this one!

PR #105, opened 2019-12-16
I487 image plus refactoring

geekscruff:
End to end demo of Image configured with M3 profile.

Various refactorings.


geekscruff:
wrong branch

PR #106, opened 2019-12-17
WIP Flexible Metadata Work So Far

geekscruff:
I don't expect you to merge it, but it will give you a sense of what work has been done.


orangewolf:
this one should be closed - both the other 2 are much further along

PR #110, opened 2020-01-16
Essi 386 solrize word boundary

rwingiupui:
Revise ALTO OCR proof-of-concept to ALTO 3.0, and hitting solr instead of fedora


aploshay:
This had a messy commit history, so I cleaned it up a bit, then added one more commit on the end.  The original history is available at `aploshay/ESSI-386_solrize_word_boundary_original`.

PR #114, opened 2020-02-04
[ESSI-442] Add collection and campus information 

andjsmit:
Caveats:

1. Only concerned with first collection found for a work or file set.
2. Campus info only added to banner if configured.


rwingiupui:
+1 Desk checked - working as designed

PR #129, opened 2020-03-06
[ESSI-751] Move extensions initializers to after_initialize

aploshay:
The extensions of `CollectionPresenter` and `WorkShowPresenter` were loading [Hyrax constants set to I18n lookup values](https://github.com/samvera/hyrax/blob/320512447cb48b2bfe6fb9a05b99d0fce193d810/app/models/hyrax/collection_type.rb#L12-L16), before the entire `I18n.load_path` was populated, resulting in "translation not found" results.

This PR resolves this by moving _all_ of the extension monkeypatches into an `after_initialize` callback, relocating these I18n lookups to later in the initialization sequence, after `locale` files for all the gems have been populated into `I18n.load_path`.

### DIFF
--- a/config/initializers/extensions.rb
+++ b/config/initializers/extensions.rb
@@ -1,29 +1,29 @@
-# active fedora node cache initialization
-ActiveFedora::Orders::OrderedList.prepend Extensions::ActiveFedora::Orders::OrderedList::InitializeNodeCache
-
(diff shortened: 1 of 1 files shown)


aploshay:
We should confirm this doesn't break anything that needs to be _before_ the application initialization

PR #132, opened 2020-03-19
ESSI-741 Add ability to optionally create BagIt bags for works

randalldfloyd:
This PR adds several components to enable creation of BagIt bags for works.  It adds:

- The bagit gem to ease creation of valid bags
- An ActiveJob job that can be initiated from various places (i.e. a rake task, a workflow method, actor stack, etc.)
- A method that can be added to workflow definitions and so that different workflows can call it whenever is appropriate (or not).


app/jobs/bag_work_job.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def perform(identifier, dropbox_path = nil, bags_path = nil, bag_info = {})
+    if dropbox_path.nil?
+      dropbox_path = File.join(ESSI.config.dig(:essi, :bagit, :default_upload_dropbox), identifier)
+    end
+    return unless Dir.exist?(dropbox_path)
+    bags_path = ESSI.config.dig(:essi, :bagit, :default_bagit_dir) unless bags_path
+    bag_path = File.join(bags_path, identifier)
+    # Check for existence of bag and overwrite
+    FileUtils.rm_rf(bag_path) if File.exist?(bag_path)
```
dlpierce: Is using `rm_rf` necessary? I'd like to see either some additional path checks and/or use of just `rm`.

randalldfloyd: Did a bunch of refactoring to rely on bagit gem methods to inspect bag files and iteratively remove vs. wholesale unsafe directory removal.  Also added tests for actually overwriting a bag with different files, which is the use case we want for works that might go back through file ingest workflow.

PR #134, opened 2020-03-30
Move primary fields outside of after init.

dlpierce:
Fixes `/app/app/forms/hyrax/bib_record_form.rb:9:in `<class:BibRecordForm>': undefined method `primary_fields=' for Hyrax::BibRecordForm:Class (NoMethodError)`


aploshay:
+1 by inspection

PR #135, opened 2020-03-30
Essi 745 bulkrax

geekscruff:
You probably don't want to merge this (especially as I broke the build) but this PR:

* adds the Bulkrax gem, using a branch containing current fixes
* files and changes added from running the Bulkrax generator

Issues:

Couldn't find collection in Export:
* Bulkrax uses Hyrax collection searching which requires a whole word (despite suggesting it works on 2 characters) - this could be improved in Hyrax

Couldn't find AdminSet in Export:
* Collection only, admin sets aren't included

Out of Date documentation
* fixed in Bulkrax

Import not working:

1) source_identifier field was not in export.csv 
* this isn't a bug per se, as the documentation for import is clear on this being required but it is a fair assumption that round-tripping would work (ie. export and import the same data) so I have made a change to include source identifier
* fixed in latest Bulkrax

2) bug in export meant exported fields were not matched to correct columns (therefore title was reported as blank and import could not proceed)
* fixed in latest Bulkrax

Browse Everything panel not loading
* ESSI configuration, not an issue with bulkrax Bulkrax
* (temporarily) commented out the s3 block in config/essi_config.docker.yml and replace with local filestore config and rebuilt the docker image


wang58:
+1. Importer and Exporter worked fine.

randalldfloyd:
Recommend closing and replacing with #152 that resolves conflicts, resolves failing tests and provides a base branch that IU developers have write access to.

PR #136, opened 2020-04-01
Lock UV version, include yarn.lock

dlpierce:
Somehow the automated docker build started pulling a newer version (3.0.36) of universalviewer during `yarn`. This makes some sense because yarn.lock wasn't committed, but it's not clear why it took until now to cause breakage since .36 has been out for a long time.

This error has revealed that the location of some files that we copy to public has changed in more recent versions. I also noticed that UV v4 is publishing prereleases, so we should consider targeting it in the future.


aploshay:
+1 by inspection

PR #141, opened 2020-04-13
ESSI-546 Send work actor create calls to an async job.

dlpierce:
Wraps the execution of the actor stack for creating works in a job to be performed asynchronously.

This involves rebuilding the request's Actor::Environment after serializing its 3 components so they can be stored in the job. Because the work is not created immediately, the browser is redirected to the user's works dashboard. 

Specs (such as the refactored IngestYamlJobSpec) may have to explicitly allow jobs to run if they involve creating works via the actor stack.

Updates to rspec related gems are included here, in particular switching to the webdriver gem.


aploshay:
+1 by inspection

PR #144, opened 2020-04-16
Eager load in test env to catch production errors

dlpierce:


aploshay:
+1 by inspection

I've also submitted #145 as a fix for #142 as this change didn't resolve the issue for me locally :(

dlpierce:
This would not have resolved that issue, instead it would cause an error during spec testing instead of later during the docker build.  However, it's causing some other errors after the first test is run.

aploshay:
@dlpierce Oh, by "did not resolve the issue", I meant that the tests still passed with the eager loading turned on, and production mode still crashed.

PR #145, opened 2020-04-16
[ESSI-759] [ESSI-808] Remove structure master module

aploshay:
Fixes the issues running in production mode, introduced in #142


dlpierce:
Seemed like a good idea, maybe we can revisit it.

PR #150, opened 2020-04-28
[ESSI-704] Move branding files to FileSet backing

aploshay:
Changes `CollectionBrandingInfo` from managing uploaded image files in the `/public/branding` subtree, to managing them via `FileSet` objects and retrieving their display images from the IIIF server.

The approach I took to modifying the 4 interrelated Hyrax components was to monkeypatch the controller, presenter, and form, but completely override the `CollectionBrandingInfo` model as that is changed drastically (and I originally tried it via monkeypatches, but that got messy, with the super calls.).

A few other notes:
- the `FileSet` gets created promptly as a stub and associated to the `CollectionbrandingInfo`, but its fleshing out is handled by the actor/job stack, so the images may not actually show up for awhile, until that process has completed.
  - `CollectionbrandingInfo` will update the `image_path` on-demand when the IIIF image URL actually becomes available
- In stock Hyrax, `CollectionBrandingInfo` has `width` and `height` fields, but they are never actually used; I left a comment about using dimensions other than default, but the views already enforce specific rendering dimensions
- there's another comment about whether to tweak the default visibility permissions on the `FileSet`
- I'm missing controller/feature coverage of all the actions to interact with adding/removing/updating `CollectionBrandingInfo` objects -- but then, that coverage wasn't present in stock Hyrax, in the first place

### DIFF
--- a/app/models/collection_branding_info.rb
+++ b/app/models/collection_branding_info.rb
@@ -1,19 +1,26 @@
-# Imported from Hyrax to modify for FileSet-backed branding images
+# Imported from Hyrax and modified for FileSet-backed branding images
 class CollectionBrandingInfo < ApplicationRecord
 
   def initialize(collection_id:,
                  filename:,
                  role:,
                  alt_txt: "",
-                 target_url: "")
+                 target_url: "",
+                 local_path: nil,
+                 file_set_id: nil,
+                 image_path: nil)
 
     super()
     self.collection_id = collection_id
     self.role = role
     self.alt_text = alt_txt
     self.target_url = target_url
+    self.file_set_id = file_set_id
+    self.image_path = image_path
 
-    self.local_path = find_local_filename(collection_id, role, filename)
+    # the file is no longer stored at this location, but this is retained
+    # for infrastructure for form field identifiers via local_path value
+    self.local_path = local_path || find_local_filename(collection_id, role, filename)
   end
 
   def collection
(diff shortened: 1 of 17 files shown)


app/models/collection_branding_info.rb
```diff
@@ -0,0 +1,116 @@
...
+    def find_local_filename(collection_id, role, filename)
+      local_dir = find_local_dir_name(collection_id, role)
+      File.join(local_dir, filename)
+    end
+  
+    def find_local_dir_name(collection_id, role)
+      File.join(Hyrax.config.branding_path, collection_id.to_s, role.to_s)
+    end  
+    
+    # FIXME: use different image dimensions than default?
```
dlpierce, line 93: Does specifying size as `full` (deprecated in iiif 3.0) or `max` work here to avoid any scaling?

aploshay:
@dlpierce @andjsmit @randalldfloyd Okay, I've force-pushed to rebase on the latest master (leaving the original commits otherwise unchanged), and added a few commits.  Of note:
- I caught and fixed leaving a FileSet orphaned after a CollectionBrandingInfo deletion; I missed this on the first pass because Hyrax handles this via controller action instead of the model.  Also added convenience methods to get a CollectionBrandingInfo from the FileSet, and see check if the FileSet is used for branding
- refactored the FileActor spec in place, prior to modification
- tweaked the FileActor to skip characterization & derivative creation for branding FileSets -- turns out, we just need the Versioning.  Also note that we had a store_files? check for CharacterizeJob in the `ingest_file` action, but not the corresponding call in the `revert_to` action, but it seemed like it should be in both places so I put the `characterize_files?` check in both.

With this, I _think_ it's ready for final review.  @dlpierce the size is getting specified as 'full' in the `image_path` so I think we're okay there

PR #151, opened 2020-04-30
Use to_prepare to allow extension reloading in dev.

dlpierce:
The various includes and prepends used in extensions do not reload correctly in dev. This moves their loading to the to_prepare hook as suggested here:
https://bibwild.wordpress.com/2016/12/27/a-class_eval-monkey-patching-pattern-with-prepend/
This fixes various errors encountered in development after making a change on a running server.


aploshay:
+1 by inspection

dlpierce:
This seemed to work really well in the dev docker-compose environment, but it would be good to test in another environment, especially if someone has a set of actions they do that cause these kind of errors and see if this fixes it.

PR #152, opened 2020-04-30
Adding Bulkrax gem and installation from generators

randalldfloyd:
This PR adds the [Bulkrax gem](https://github.com/samvera-labs/bulkrax) into the ESSI code base. It is a replacement of #135 that resolves conflicts, resolves failing tests and provides a base branch that IU developers have write access to.


randalldfloyd:
Ying confirmed that this works locally with `essi_816_bulkrax` branch.

PR #154, opened 2020-04-30
Essi-668 cantaloupe support and docker fixes

dlpierce:
Primary goal here is to run cantaloupe for IIIF. Along the way came many related fixes for the docker-compose environment.

- Cantaloupe provides its IIIF API at `/iiif/2` while riiif by default uses `/images`. To allow seamless failover, and to avoid unnecessarily complex proxy configuration, riiif is now mounted at `/iiif/2` and should takeover when a proxy is not in place to direct requests to Cantaloupe.
- Thumbnail url generation should now be based on the thumbnail_id (hasRelatedImage_ssim in solr) and not on thumbnail_path. This should progress ESSI-822 as well.
- Cantaloupe delegate script is updated to use environment variables for fedora config.
- A sample nginx proxy config is provided and used in docker-compose.
- Nodejs is updated to version 12 (from 8) to fix missing shx.
- The docker-compose essi_config is loaded via a volume to avoid rebuilding the image after changes.
- The fcrepo docker data volume location is corrected.
- Essi docker images do not run as root. Better for security and avoids writing root owned asset files to the app volume.


dlpierce:
helium/essi-stage will need configuration updates when this is merged.

aploshay:
:+1: by inspection

dlpierce:
I'm reworking the UID changes.

app/helpers/catalog_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+  include Blacklight::CatalogHelperBehavior
+
+  ##
+  # Get the URL to a document's thumbnail image
+  # Overrides Blacklight behaviour
+  #
+  # @param [SolrDocument, Presenter] document
+  # @return [String]
+  def thumbnail_url document
+    representative_document = ::SolrDocument.find(document.thumbnail_id)
```
aploshay, line 12: I think the first two lines need an `if` check, or equivalent, against `thumbnail_id` being `nil` -- I just created a new Collection, without any works assigned to it and no representative thumbnail, and hit this condition on the front page

dlpierce: Refactored this to show the default image if anything fails.

PR #155, opened 2020-05-01
Fix assets:precompile in prod mode

dlpierce:
Bulkrax JS additions broke assets:precompile in prod. This updates our config to match https://github.com/lautis/uglifier#rails


aploshay:
+1 by inspection

PR #158, opened 2020-05-12
[ESSI-822] Revise collection thumbnails

aploshay:
The first commit uses IIIF Thumbnails for both collection types (User Collections, Admin Sets), as is already done for works.

The second one tweaks `#thumbnail_url` logic in `CatalogHelper`, but I'm less confident it's the right change.  The overall context for this, is: Hyrax uses a mix of two paradigms for displaying thumbnails:
* **Blacklight-based**, doing a `render_thumbnail_tag` call on a `SolrDocument` _document_ record
* **Presenter-based**, calling `thumbnail_path` on the _presenter_
This is why, for objects with no `thumbnail_id`, we got a mix of the `default.png` display of a dogeared page (when using the Blacklight methodology) that's agnostic of the object's class, versus displaying generic 'work.png' (a single box) or 'collection.png' (a stack of 3 boxes) -- the indexers for works and collections know to default to those more specific generic images, instead of `default.png`  (Note that in all 3 of these cases, the `thumbnail_url` returned is a raw link to public assets, rather than the IIIF server.  Presumably that's okay?)

@dlpierce This probably merits your review as it's touching thumbnail content you've worked on quite recently.

### DIFF
--- a/app/helpers/catalog_helper.rb
+++ b/app/helpers/catalog_helper.rb
@@ -9,6 +9,7 @@ module CatalogHelper
   # @param [SolrDocument, Presenter] document
   # @return [String]
   def thumbnail_url document
+    return document.thumbnail_path if document.try(:thumbnail_path).present?
     if document.id == document.thumbnail_id
       representative_document = document
     else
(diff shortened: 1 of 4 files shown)


aploshay:
Removing the WIP since the change is okay with @dlpierce

PR #159, opened 2020-05-13
[ESSI-831] Turn source_metadata_identifier errors into alerts

aploshay:
The remote source metadata logic we pulled from pumpkin was missing error handling for two cases:
* supplying an invalid `source_metadata_identifier`
* attempting to pull remote metadata and getting no results for the identifier

This PR handles both cases, but whereas in pumpkin these cases blocked saving the work and flashed a _failure_ message to the user, in ESSI both cases succeed and merely flash an _alert_ to the user.  The validation logic for a `source_metadata_identifier` is also loosened to allow periods and hyphens.

### DIFF
--- a/app/controllers/concerns/essi/remote_metadata_lookup_behavior.rb
+++ b/app/controllers/concerns/essi/remote_metadata_lookup_behavior.rb
@@ -16,6 +16,7 @@ module ESSI
     end
 
     def remote_attributes
+      return {} if remote_data.nil?
       @remote_attributes ||= begin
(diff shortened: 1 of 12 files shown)


.solr_wrapper
```diff
@@ -1,5 +1,5 @@
 # Place any default configuration for solr_wrapper here
-# version: 7.7.1
+version: 7.7.1
```
dlpierce, line 2: Is this meant to be included here?

PR #161, opened 2020-05-18
WIP [ESSI-763] Image QC libraries for uploaded files

aploshay:
Early post of a work-in-progress.  The first two commits are the actual libraries, which may be used in console.  The third commit adds an actor to allow using the qc stack in the UI, for the sake of convenience -- but long-term it's meant to be removed, and QC applied through other means (such as workflow, or possibly in BrowseEverything at time of upload, although it may not performant enough for that application to be practical.)

Overview of the current version of the proposed framework:
* a source-agnostic **metadata reader layer** provides image metadata
* a configurable **validation layer** runs **validation rules** against the available metadata

More detail:
* Metadata reader layer:
  * This is currently provided by classes that read from a MiniMagick::Image, but could also be a simple wrapper around metadata provided by other sources.  (For example, given a Hash of metadata from some 3rd-party tools, the wrapper could pass along method calls as `#fetch` calls to the Hash.)
  * there's support right now for running MiniMagick on either :imagemagick or :graphicsmagick CLI backends, but the :graphicsmagick output is a little incomplete and inconsistent compared to :imagemagick, and there's no support yet for an :imagemagick7 backend
* Validation layer:
  * Exists with the expectation that metadata properties (whether simple, or complex) are provided by the metadata reader layer, and this just applies validation rules against them
  * a validator has a set of default rules which may be extended/overriding by initializing with any new/modified rules
  * the validator can apply simple rules without needing any new methods
    * custom methods can be added when needed; the [IUB validator provides examples of custom methods](https://github.com/aploshay/essi/blob/ESSI-763_file_qc/lib/image_qc/image_validator/iub_image_validator.rb#L17-L29)
  * a validator could also bypass use of the metadata reader layer entirely, and pull validation results directly from some other source, such as ImageProc output files
* Validation rules:
  * exist as a Hash, which may be manually specified, sourced from a YAML file, etc.
  * are [documented along with a sample set in the CustomizableImageValidator](https://github.com/aploshay/essi/blob/ESSI-763_file_qc/lib/image_qc/image_validator/customizable_image_validator.rb#L6-L32), and other working examples are provided in the [IUBImageValidator](https://github.com/aploshay/essi/blob/ESSI-763_file_qc/lib/image_qc/image_validator/iub_image_validator.rb#L4-L15) and [IUPUIImageValidator](https://github.com/aploshay/essi/blob/ESSI-763_file_qc/lib/image_qc/image_validator/iupui_image_validator.rb#L4-L7)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ RUN groupadd -g ${GROUP_ID} essi && \
     apt-get update -qq && \
     apt-get install -y build-essential default-jre-headless libpq-dev nodejs \
       libreoffice-writer libreoffice-impress imagemagick unzip ghostscript && \
+# FIXME: install graphicsmagick, exiftool?
     apt-get install -y libtesseract-dev libleptonica-dev liblept5 tesseract-ocr -t testing \
     yarn && \
     rm -rf /var/lib/apt/lists/*
--- a/app/actors/essi/actors/quality_control_uploaded_files_actor.rb
+++ b/app/actors/essi/actors/quality_control_uploaded_files_actor.rb
@@ -1,0 +1,58 @@
+module ESSI
+  module Actors
+    class QualityControlUploadedFilesActor < Hyrax::Actors::AbstractActor
+      # @param [Hyrax::Actors::Environment] env
+      # @return [Boolean] true if create was successful
+      def create(env)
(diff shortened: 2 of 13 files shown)


aploshay:
Closing, linking to from ESSI-755 for reference

PR #165, opened 2020-05-26
[ESSI-761] Apply jp2 image conversion, if configured

aploshay:
Some notes:
* I took out the kakadu Docker stuff, and [stashed it in a gist](https://gist.github.com/aploshay/10e4fa4c410c00577bd9b80b7b70f6a4).
* The ingest logic now attempts to infer the `source_metadata_identifier` value from a `FileSet` from the uploaded filename
* This also includes a change to the logic on storing a link to preservation masters -- whether to do it, and how.  
  * Previously, this was an either/or against the config value for `store_original_files` -- now, it looks at whether a value is present for `master_file_service_url`, and infers the presence of a value to indicate storing a link to masters.  Consequences are:
    * You'll get empty `FileSet` objects unless you set at _least_ one of `store_original_files`, `master_file_service_url`; presumably you don't want that, and should set either, or both
    * If storing a link to the master, the file use is still `original_file` if we're not storing original files, _or_ `preservation_master_file` if we are storing both original files and masters.  (Consistently storing the link as `preservation_master_file` is problematic because there are baked-in assumptions that a `FileSet` has an `original_file` -- so it switches to that use when `store_original_files` is false.)

### DIFF
--- a/app/actors/hyrax/actors/file_actor.rb
+++ b/app/actors/hyrax/actors/file_actor.rb
@@ -22,16 +22,21 @@ module Hyrax
       # @see IngestJob
       # @todo create a job to monitor the temp directory (or in a multi-worker system, directories!) to prune old files that have made it into the repo
       def ingest_file(io)
-        # FIXME: pass along mimetype, filename options?
+        infer_source_metadata_identifier(file_set, io)
         if store_files?
+          transform_to_jp2(io) if transform_to_jp2?
+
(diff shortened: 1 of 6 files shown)


dlpierce:
Separate lines is fine here with the backslashes. We might want to consider moving off of the stretch base image so we can stop using things from the stretch testing repository.

PR #177, opened 2020-06-05
essi-910 Generate pdfs 

ShanaLMoore:
JIRA: https://bugs.dlib.indiana.edu/browse/ESSI-910

**//TODO:**

- [ ] Handle image size
- [ ] Create a service and specs for generating the pdf. The service will pass the path of the generated pdf to the pdf endpoint of the paged_resource controller. Use a redirect for the hyrax downloads path to render the pdf for download in the browser.

### DIFF
--- a/app/controllers/hyrax/paged_resources_controller.rb
+++ b/app/controllers/hyrax/paged_resources_controller.rb
@@ -22,24 +22,27 @@ module Hyrax
       pdf_hash = generate_pdf(resource)
 
       send_file pdf_hash[:path],
-        filename: pdf_hash[:file_name],
-        type: 'application/pdf',
-        disposition: 'inline'
+                filename: pdf_hash[:file_name],
+                type: 'application/pdf',
+                disposition: 'inline'
     end
 
     private
 
+    # TODO: extract into service
     def generate_pdf(resource)
-      path = Rails.root.join('tmp', 'pdfs')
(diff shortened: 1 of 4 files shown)


app/controllers/hyrax/paged_resources_controller.rb
```diff
@@ -16,5 +16,27 @@ class PagedResourcesController < ApplicationController
 
     # Use this line if you want to use a custom presenter
     self.show_presenter = Hyrax::PagedResourcePresenter
+
+    def pdf
+      # PagedResourcePDF.new(presenter, quality: params[:pdf_quality]).render(pdf_hyrax_paged_resource_path)
+      # redirect_to main_app.download_path(presenter)
+
+      resource = PagedResource.find(params[:id])
+      # TODO: if tmp/pdfs/ doesn't exist, mkdir it
```
ShanaLMoore, line 25: you could try:

```FileUtils.mkdir_p(Rails.root.join('tmp', 'pdfs'))```

aploshay: Pertaining to another story, @dlpierce indicated that for our enterprise Docker context at IU, using the system temp dir may be preferable, as ImportUrlJob does:
https://github.com/samvera/hyrax/blob/master/app/jobs/import_url_job.rb#L71-L85

PR #178, opened 2020-06-08
Use public url method for accessing S3/minio and pin minio image version

dlpierce:
This avoids issues with signed urls expiring or not being processed correctly in browse everything.
This requires setting the bucket in minio to public using the mc utility. Current verion of minio is not showing the nas gateway bucket in mc, so an older version has been pinned.


aploshay:
Okay, confirmed that Upload from Server using the S3 source still works.

PR #179, opened 2020-06-10
ESSI-917 truncate collection description/exempt HTML string

wang58:
This is a subtask of ESSI-879 (ESSI-879 has been merged to master) This change is exempt HTML String from truncating to 160 chars.


wang58:
@dlpierce thanks for your comment. But I have two concerns here. 1)using CSS property text-overflow is hard to track 160 characters (the width property can not be the char number.) 2) this truncate method is borrowed from the sub-collection description truncate. I had a discussion with Nick and Adam, that's what they want.

dlpierce:
Okay, I think this will work for now. There is an issue where html special characters will be escaped in truncated text (& becomes &amp;). I might come back to this and see if the logic can be moved to a helper method.

PR #182, opened 2020-06-11
WIP Bundle updates

aploshay:
To placate dependabot; required bumping minor versions of hyrax and rails, which merits testing I have not done.
UPDATE: I think we're in dependency hell with active-fedora :(


aploshay:
I think we'll have to do a fuller hyrax upgrade to satisfy dependabot.

PR #184, opened 2020-06-11
ESSI-899 PagedResources PDF download

ShanaLMoore:
JIRA: https://bugs.dlib.indiana.edu/browse/ESSI-899

DEMO: https://share.getcloudapp.com/P8uel79K


.gitignore
```diff
@@ -46,3 +46,7 @@ staged_files
 
 public/assets
 vendor/engines/bulkrax/
+spec/fixtures/pdfs
```
orangewolf, line 49: if these are fixtures, why are they ignored?

docker-compose.yml
```diff
@@ -38,18 +38,21 @@ services:
     ports:
       - "80:8980"
     depends_on:
-      - app
+      - essi_app
       - worker
 
-  app:
+  essi_app:
```
orangewolf, line 44: this feels like an arbitrary change?

docker-compose.yml
```diff
@@ -38,18 +38,21 @@ services:
...
     << : *default-app
     command: bash -l -c "rake db:migrate db:seed && yarn install && bundle exec puma -b tcp://0.0.0.0:3000"
     environment:
       VIRTUAL_HOST: essi.docker
       VIRTUAL_PORT: 3000
       IN_DOCKER: 'true'
+      HUB_URL: http://chrome:4444/wd/hub # <-- Add the HUB_URL environment variable
```
orangewolf, line 51: I'm not sure this code comment adds info. it kinda says "this thing that is right here is here"

docker-compose.yml
```diff
@@ -145,8 +148,13 @@ services:
       - "8080"
 
   chrome:
-    image: selenium/standalone-chrome:3.141
-    shm_size: 2G
+    image: selenium/standalone-chrome-debug:3.141.59-20200525
+    volumes:
+      - /dev/shm:/dev/shm
+    shm_size: 2
```
orangewolf, line 154: bring back the G, it doesn't affect much, but is more clear.

docker-compose.yml
```diff
@@ -145,8 +148,13 @@ services:
...
   chrome:
-    image: selenium/standalone-chrome:3.141
-    shm_size: 2G
+    image: selenium/standalone-chrome-debug:3.141.59-20200525
+    volumes:
+      - /dev/shm:/dev/shm
+    shm_size: 2 
+    ports:
+      - "4444:4444"
+      - "5959:5900"
```
orangewolf, line 157: do you need these port bindings

essi-nginx.nginx
```diff
@@ -1,5 +1,5 @@
 upstream essi_app {
-  server app:3000;
+  server essi_app:3000;
```
orangewolf, line 2: app

spec/rails_helper.rb
```diff
@@ -43,25 +55,23 @@
   TEST_HOST='essi.docker'
   capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
     chromeOptions: {
-      args: %w[headless disable-gpu no-sandbox whitelisted-ips window-size=1400,1400]
+      args: %w[disable-gpu no-sandbox whitelisted-ips window-size=1400,1400] #run without headless
```
orangewolf, line 58: can you expand this comment with the why? #run without headless so we can see screenshots (for example)

spec/rails_helper.rb
```diff
@@ -11,14 +11,26 @@
...
 require 'capybara/rails'
+require 'capybara-screenshot/rspec'
 require 'database_cleaner'
 require 'active_fedora/cleaner'
 require 'selenium-webdriver'
 require 'webdrivers'
 require 'i18n/debug' if ENV['I18N_DEBUG']
 require 'byebug' unless ENV['CI']
 require 'noid/rails/rspec'
+# without lines 23-32, screenshots and html captured from failing specs are blank
```
orangewolf, line 22: will this work if you do it in a support file? you include the supports next, so I bet it will and then it doesn't need to clutter up rails_helper

spec/rails_helper.rb
```diff
@@ -82,7 +92,48 @@
 Capybara.default_driver = :rack_test # This is a faster driver
 Capybara.javascript_driver = :selenium_chrome_headless_sandboxless # This is slower
 
+Capybara::Screenshot.register_driver(:selenium_chrome_headless_sandboxless) do |driver, path|
```
orangewolf, line 95: I belong in a support file please

spec/rails_helper.rb
```diff
@@ -82,7 +92,48 @@
...
+  page.save_screenshot(screenshot_path)
+
+  page_name = "html-#{filename}-#{line_number}-#{timestamp}.html"
+  page_path = "#{Rails.root.join('tmp/capybara')}/#{page_name}"
+  page.save_page(page_path)
+
+  puts "\n  Screenshot: tmp/capybara/#{screenshot_name}"
+  puts "  HTML: tmp/capybara/#{page_name}"
+end
+
```
orangewolf, line 121: end stuff that belongs in a support file

spec/rails_helper.rb
```diff
@@ -82,7 +92,48 @@
...
 RSpec.configure do |config|
+  config.before(:each, type: :system) do
+    driven_by :rack_test
```
orangewolf, line 124: are you sure this isn't the default?

spec/rails_helper.rb
```diff
@@ -82,7 +92,48 @@
...
+  end
+
+  config.before(:each, type: :system, js: true) do
+    driven_by :selenium_chrome_headless_sandboxless #
+
+    if ENV['IN_DOCKER'].present?
+      Capybara.server_host = '0.0.0.0'
+      Capybara.server_port = 3010
+      # renaming container because the app domain is taken by google(gTLD)
+      Capybara.app_host = 'http://essi_app:3010'
```
orangewolf, line 134: ah! I take it all back, that makes sense.  maybe just call it "essi" everywhere instead of essi_app

orangewolf: if this config is done above (where it is in hyku and where you deleted it) it should work just fine. I don't think you need to set it every single time like this.

spec/rails_helper.rb
```diff
@@ -110,7 +161,7 @@
   config.filter_rails_from_backtrace!
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
-
+  config.include FactoryBot::Syntax::Methods
```
orangewolf, line 164: I think this line is un-needed if you have factory bot rails isn't it?

spec/services/essi/generate_pdf_service_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'rails_helper'
+
+
+RSpec.describe ESSI::GeneratePdfService do
+  let(:resource) { create(:paged_resource) }
+  let(:service) { described_class.new(resource) }
+  let(:pdf) { service.generate }
+
+  before do
+    pdf_fixtures = Rails.root.join('spec', 'fixtures', 'pdfs')
```
orangewolf, line 10: these should be put in tmp if they are tmp files, not in spec/fixtures. then they don't need to be gitignored separately.

.gitignore
orangewolf: (see below)

docker-compose.yml
orangewolf: I see - this is about google having app as a domain. maybe just "essi" is a better name?

docker-compose.yml
ShanaLMoore: These port bindings let me use screen sharing tool to watch the specs go through the ui.

essi-nginx.nginx
ShanaLMoore: renamed to essi

spec/rails_helper.rb
ShanaLMoore: Without this, I have to specify FactoryBot instead of just calling create within the rspec let blocks.

spec/rails_helper.rb
ShanaLMoore: Should I not just put all of the capybara set up into its own file?

spec/rails_helper.rb
ShanaLMoore: yep, it is indeed.

docker-compose.yml
orangewolf: that sounds pretty cool. maybe a corresponding snippet in the readme?

spec/rails_helper.rb
orangewolf: yes, that is what I was suggesting and would have been clear if I was more interested in communicating and less interested in sounding clever.  Sorry =-(

spec/rails_helper.rb
orangewolf: gotcha

docker-compose.yml
```diff
@@ -38,18 +38,21 @@ services:
...
     command: bash -l -c "rake db:migrate db:seed && yarn install && bundle exec puma -b tcp://0.0.0.0:3000"
     environment:
       VIRTUAL_HOST: essi.docker
       VIRTUAL_PORT: 3000
       IN_DOCKER: 'true'
+      HUB_URL: http://chrome:4444/wd/hub 
     ports:
       - "3000:3000"
+    depends_on:
+      - chrome
```
dlpierce, line 55: This will overwrite the depends_on key instead of append to the values coming from default-app. It should either be removed and started manually when needed or moved to the list in default-app.

essi-nginx.nginx
```diff
@@ -25,7 +25,7 @@ server {
     proxy_set_header        Connection $connection_upgrade;
 
     # Fix the "It appears that your reverse proxy set up is broken" error.
-    proxy_pass          http://essi_app;
+    proxy_pass          http://essi;
```
dlpierce, line 28: This looks like a hostname, but it is actually referring to the upstream defined above.
```suggestion
    proxy_pass          http://essi_app;
```

docker-compose.yml
ShanaLMoore: Thanks for the catch, I will remove it.

ShanaLMoore:
> The docker related changes seem fine to me. EDIT: Actually, see my request below.
> Are there gems that provide suitable font files that we could use instead of directly including them here?
> And is the icomoon font actually being used? I didn't see any reference to it.

The font files are copied over from pumpkin, which are used to generate the pdf cover page. We copied them over directly for consistency's sake, but can explore other options/fonts/gems if preferred.

PR #185, opened 2020-06-17
Essi-357 Use buster docker image, save dev rails cache 

dlpierce:
This uses the debian buster image as the base for the essi docker image. Previously it was pinned to stretch because IU CAS SSL was incompatible.

This also adds an enhancement to the dev docker compose stack that saves the rails application cache between restarts.


orangewolf:
👍 the cache will make every page load faster in docker for mac. the file system sync does not like lots of small changing files and the assets cache is by full of lots of small changing files.

PR #186, opened 2020-06-22
Essi-952 Upgrade UV and disable thumbnail cache invalidation.

dlpierce:
Updates Universal Viewer to 3.1.1 and disables the adding of timestamps to thumbnail image requests that were causing the cantaloupe to miss its cache.

We may want to review the Cache-Control headers sent in the image response as it currently has max-age=2592000 (30 days).

This PR also consolidates the two uv_config.json files we had, however some of these options may now be outdated as they seem to not have an effect (no "TOC" for the index panel) or cause layout issues (image selection header). **Update: These options have been removed from this PR.**

uv.html has been updated with the upstream version.


aploshay:
Looks good, runs fine locally; confirming in Docker.

PR #187, opened 2020-06-22
Essi 913 pdf download setting

labradford:
Add a setting to the ESSI config to turn on/off pdf downloads
Make sure current user is admin before showing pdf download button

Pull Request https://github.com/IU-Libraries-Joint-Development/essi/pull/184 must be merged before this PR

Both conditions met:
image

One or both conditions not met:
image

### DIFF
--- a/.env
+++ b/.env
@@ -5,3 +5,4 @@ RAILS_LOG_TO_STDOUT=true
 REGISTRY_HOST=registry.gitlab.com
 REGISTRY_URI=/notch8/essi
 TAG=dev
+ALLOW_PDF_DOWNLOAD=true
--- a/app/presenters/hyrax/paged_resource_presenter.rb
+++ b/app/presenters/hyrax/paged_resource_presenter.rb
@@ -23,7 +23,7 @@ module Hyrax
       renderings = []
       if rendering_ids.present?
         rendering_ids.each do |file_set_id|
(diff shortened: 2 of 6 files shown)


.env
```diff
@@ -5,3 +5,4 @@ RAILS_LOG_TO_STDOUT=true
 REGISTRY_HOST=registry.gitlab.com
 REGISTRY_URI=/notch8/essi
 TAG=dev
```
ShanaLMoore, line 7: Should we add the .env to .gitignore? I realize it's already in this repo 🤔

app/presenters/hyrax/paged_resource_presenter.rb
```diff
@@ -23,7 +23,7 @@ def sequence_rendering
       renderings = []
       if rendering_ids.present?
         rendering_ids.each do |file_set_id|
-          renderings << manifest_helper.build_pdf_rendering(file_set_id)
+          renderings << manifest_helper.build_pdf_rendering(file_set_id) if ESSI.config.dig(:essi, :allow_pdf_download) && current_ability.current_user.admin?
```
ShanaLMoore, line 26: @labradford Actually, maybe we should add this check to the controller #pdf method too, in case another clever non admins tries to go around the download button.

labradford: added check to pdf method

PR #199, opened 2020-07-29
ESSI 465 Flexible Metadata Gem Version

orangewolf:


orangewolf:
need to work through the conflicts and remove a few things I changed just to get the app working locally

orangewolf:
Pushed an update here that fixes issue we were seeing with the admin set going missing during the spec runs. still having issues running this app with the current docker config on our machines, but I believe this will be closer to a "real" number of failures than it was before.

randalldfloyd:
Superseded by #233

PR #200, opened 2020-07-29
Optimize image used for OCR with textcleaner and bitonal conversions.

andjsmit:


rwingiupui:
+1 - Tested locally, appears to be working, no issues found.

bin/textcleaner
```diff
@@ -0,0 +1,619 @@
...
+# -e      enhance           enhance image brightness before cleaning;
+#                           choices are: none, stretch or normalize; 
+#                           default=stretch
+# -f      filtersize        size of filter used to clean background;
+#                           integer>0; default=15
+# -o      offset            offset of filter in percent used to reduce noise;
+#                           integer>=0; default=5
+# -u                        unrotate image; cannot unrotate more than 
+#                           about 5 degrees
+# -t      threshold			text smoothing threshold; 0<=threshold<=100;
```
dlpierce, line 57: It looks like the textcleaner script has builtin support for the threshold option. Is there a reason for performing it as an separate additional step?

bin/textcleaner
```diff
@@ -0,0 +1,619 @@
...
+if [ "$enhance" = "stretch" ]; then
+	enhancing="$setcspace -contrast-stretch 0"
+elif [ "$enhance" = "normalize" ]; then
+	enhancing="$setcspace -normalize"
+else
+	enhancing=""
+fi
+#echo "enhancing=$enhancing"
+
+# setup blurring
```
aploshay, line 533: @andjsmit @dlpierce It looks like the "threshold" argument to textcleaner is massaged into a "blurring" argument actually passed to `convert`; I don't see that `textcleaner` actually has a way to set a real `threshold` value for `convert`

dlpierce: Okay then. Maybe there's some later optimization that could be done by rolling that option into the textcleaner script.

PR #207, opened 2020-08-14
Adds checks for running an access-only site to hide controls and override routes

randalldfloyd:
This is a conservative first pass to hide and disable privileged controls when the site is access-only.  It provides checks for hiding CRUD controls and for rerouting all requests for the dashboard to to the homepage (it also hides the dashboard links from the user dropdown menu.)

The checks are only activated in the presence of the config setting `site_usage= access_only`.  The setting does not have to exist in the config and absence of it will imply a cataloging site with full controls.

It does not cut into ability and does not disable resourceful routes to work type concerns so it would still be possible to craft CRUD URLs, but unlikely, especially in the absence of the dashboard.  This is at least one big step towards access-only compared to current legacy Pages where all controls still exist in the access site. 

Follow-on stories could explore disabling ability and overriding resourceful routing for work type concerns.


app/views/hyrax/base/_show_actions.html.erb
```diff
@@ -18,7 +18,7 @@
       <% end %>
     </div>
     <div class="col-sm-8 text-right">
-    <% if presenter.editor? %>
+    <% unless presenter.editor? && ESSI.config.dig(:essi, :site_usage) == 'access_only' %>
```
dlpierce, line 21: I'd like to see the site_usage check as a helper method.

config/routes.rb
```diff
@@ -11,6 +11,26 @@
         mount BrowseEverything::Engine => '/browse'
 
   mount Riiif::Engine => 'iiif/2', as: :riiif if Hyrax.config.iiif_image_server?
+
+  # Hide top-level routes to the dashboard when running the public access site.
```
dlpierce, line 15: Should these redirects be the first routes defined so they are matched first? I'd worry that Bulkrax, BrowseEverything, etc might expose a route.

randalldfloyd: Yes to that. Routes moved to very top and still function correctly. Notice I didn't use the suggested helper in the route logic - I'm unsure of the class loading order and whether routes might be drawn before the helper can be resolved.

app/views/hyrax/base/_show_actions.html.erb
randalldfloyd: Suggested helper was placed in ApplicationHelper because two completely different controller contexts need it: work show and the user profile dropdown.

dlpierce (approved):

PR #208, opened 2020-08-14
[ESSI-1013] Delivering .jp2 mimetype to Cantaloupe

aploshay:
Currently in draft status, pending confirmation of effectiveness and desirability of the approach.

For the case of a .tiff upload transformed to .jp2, uses the .jp2 for file characterization, which correctly sets the mimetype.

Continues to use originally-uploaded .tiff file for derivative creation.

Considerations:
* The current branch does not yet go back so set a .tiff mimetype for the FileSet, itself -- or doing that in some TBD other location
* We _probably_ want to characterize the _actual_ file we have and store, since that's what we can later run fixity checks against?
  * If we also want to characterize the original TIFF file, we'd need another location to store it.


randalldfloyd:
@aploshay This is looking good, I need to desk check it.  For how we treat an original TIF, we should move forward with the assumption that they will have come through a dropbox processor and already have technical metadata etc. taken for it.  It's still sketchy whether or not we have to provide upload from local machine in the initial MVP, and in that case we would be missing it for the original TIF.  In the ideal future workflow, outside processing might do the JP2 creation and inform the jobs to eliminate some work.

andjsmit (approved):
+1 - by Desk check. Imported several TIFF files which were converted to JP2s. Metadata on FileSets matched downloaded JP2 files.

PR #214, opened 2020-09-16
[ESSI-994] Collection banner descriptions

aploshay:
Re-based on updated master that included merging #215.

* I don't think that modifying `app/assets/stylesheets/hyrax.scss` with an additional line for explicitly importing the customized collections scss (`rb @import 'hyrax/collections'`) should actually be necessary, and does seem to lead to the customized css file being included _twice_ -- but without that line, the customized css doesn't seem to be included, _at all_?
* Also includes a couple of tweaks to the main banner -- removing rounded corners, removing maximum width

UPDATE: modified inclusion of tweaks to collections css


app/assets/stylesheets/hyrax/_collections.scss
```diff
@@ -0,0 +1,51 @@
...
+    .label {
+      border: 1px solid rgba(255, 255, 255, 0.5);
+      box-shadow: 1px 2px 3px rgba(0, 0, 0, 0.25);
+      margin-left: 0.5em;
+      vertical-align: 100%;
+    }
+
+  }
+
+}
```
dlpierce, line 51: Could this be moved to it's own uniquely named file to perhaps resolve the odd double inclusion behavior? It doesn't use any scss variables so I think it could be it's own css file that overrides the earlier declarations from the stock hyrax/_collections.scss

aploshay: @dlpierce Should be resolved in new commit

aploshay:
Re-based on updated master

dlpierce (approved):

PR #215, opened 2020-09-16
Essi 993 collection banner image

wang58:
There are two places to move the collection banner image up. One is for the work which belong to a collection(has a banner image). Another is for the collection( with the banner image) itself.


aploshay (approved):
+1

PR #217, opened 2020-09-24
ESSI-1081 Remove blank space between pre-footer and IU footer

wang58:
Remove the margin-top in footer css. This change will remove the blank space in between mahogany pre-footer bottom and crimson line above IU footer


rwingiupui:
Desk Check - No more blank space +1

PR #220, opened 2020-09-29
[ESSI-1086] Adds user roles depending on users's LDAP groups and ESSI configuration.

andjsmit:


rwingiupui:
+1 Desk checked - works fine.

PR #222, opened 2020-10-01
Essi 1080 last updated move

wang58:
Move "Last Updated" from banner to Collection Details list


dlpierce (changes requested):
I think this could be refactored to use the recently added lib/extensions/hyrax/collection_presenter/customized_terms.rb

wang58:
@dlpierce thanks! I have been thinking to have something like customized_term  to do this work. And thank you very much for pointing it out that we had it already.

dlpierce (approved):

PR #225, opened 2020-10-15
[ESSI-1022] Browse by campus

aploshay:
Adds to homepage grid of campus listings, to browse collections for that campus


dlpierce:
Would `app/assets/images/campuses` be better?

dlpierce (approved):

PR #226, opened 2020-10-19
ESSI-802 Run jobs depending on file_set parent later in the stack.

dlpierce:
When using the OrderedMembers flavor of actors in combination with remote file uploads, the fileset parent relationship is not always persisted when it is needed for creating certain asynchronous jobs. This queues those jobs again later in the actor stack after the work membership has been saved.

There may still be times when VisibilityCopyJob and InheritPermissionsJob are created with a nil argument, but a workaround is provided to skip those jobs in this case.

Also includes a fix for thread safety (hyrax PR #3477).


aploshay (approved):
+1 by inspection

I do like the pattern of a separate commit for a Hyrax vs its customized version, but it's also easy enough to diff our version of `environment.rb` against Hyrax's

dlpierce:
Good point. I split this into 3 commits.

randalldfloyd (approved):
This seems like a great solution and the PR is in good order.

randalldfloyd (approved):
+1 This still seems ready to merge, though I don't have time to desk-check it this morning.

PR #227, opened 2020-10-19
ESSI-1082 Move some footer info to About page

wang58:
Move Digital Collections and Samvera Hyrax version/release information from mahogany footer to "About" page


aploshay (approved):
+1

PR #231, opened 2020-11-05
essi-980: Use IIIF to convert images when generating PDF files

nianma:


dlpierce:
Closed in favor of #234

PR #232, opened 2020-11-05
[ESSI-1015] Abstract textcleaner dependency into ocr_preprocessor

aploshay:


rwingiupui:
+1 Desk Checked - Appears to be working as designed.

PR #234, opened 2020-11-05
ESSI-980 IIIF PDF

dlpierce:


aploshay:
+1 by inspection but will confirm via running it locally, directly and containerized

aploshay (approved):

PR #244, opened 2020-12-10
essi-1103: remove the description in gallery view

nianma:


app/views/catalog/_index_gallery_default.html.erb
```diff
@@ -1,12 +1,2 @@
 <% doc_presenter = index_presenter(document) %> 
 <%# default partial to display solr document fields in catalog index view -%>
```
aploshay, line 2: This comment line is no longer accurate with the block below removed; I'd advocate updating the comment line to reflect the removal, or removing the line entirely

dlpierce: Really the whole file could be blank since the first line is not outputting anything. Leaving an empty file will allow for non-default items to still show details if that's something needed in the future.

dlpierce, line 2:
```suggestion
<%# blank partial to disable details in default gallery view -%>
```

PR #245, opened 2020-12-10
essi-1167: turn off the bookmark option

nianma:


aploshay:
If we're going to override a file from Hyrax, I like following the pattern Tamsin recommended of making 2 separate commits:
- bring in the file, unmodified
- commit the modifications
It makes is easier to see what was changed, and why.

Monkeypatching would be another option.

dlpierce:
Closed in favor of #254

PR #251, opened 2020-12-17
ESSI-1178 homepage html heading change

wang58:


rwingiupui:
+1 - Desk checked - new change in place as requested.

PR #252, opened 2020-12-17
Fix port missing from final url

dlpierce:
Upstream gem this is based on ignores port when building a uri string. This uses URI to_s instead.


dlpierce:
I don't know why this gem insisted on reinventing to_s.

PR #259, opened 2021-02-05
ESSI-1230 Hiding all action buttons for works when site usage is access only

randalldfloyd:
Corrects a bug in logic that hides edit action buttons  if `site_usage == access_only` and additionally hides the remaining left side buttons for adding to collection and adding as featured item.


wang58:
Looks good to me.

PR #266, opened 2021-02-17
essi-1181: change search result title heading to h3

nianma:


rwingiupui:
+1 - Desk checked search result title headings - web browser inspector shows titles are h3.

PR #272, opened 2021-03-03
improve the accessibility for child item

nianma:


app/assets/stylesheets/hyrax/accessibility_tweaks.css
```diff
@@ -0,0 +1,4 @@
+table.related-files td.attribute-permission span.label {
```
aploshay, line 1: It might be a good pattern to have a comment line before a new CSS block indicating the story it's for?

PR #273, opened 2021-03-03
ESSI-1226 accessibility broken aria reference

wang58:
The error is due to aria-labelledby tag in "_ajax_model.html.erb'. After look into the code and check the hyrax. The file does not necessarily override. The origin "ajax_modal.html.erb' is from /gems/ruby-2.6.3/gems//blacklight-6.24.0/app/views/shared/_ajax_modal.html.erb and does not have aria-labelledby tag there.  So I think the better solution would be removing the whole file.


aploshay (approved):
+1 by inspections; looks like this was fixed on the blacklight side somewhere between 6.15 and 6.24

PR #276, opened 2021-03-08
Resolve deprecation warning for #field_value calls

aploshay:
Resolves:

DEPRECATION WARNING: You provided a String value to IndexPresenter#field_value Provide a Blacklight::Configuration::Field instead. This behavior is deprecated in Blacklight 7.


randalldfloyd (approved):
Seems to desk check okay.

PR #284, opened 2021-04-01
[ESSI-1195] Add and use homepage_banner.png

aploshay:
A few things to note:
* For this to work in production, we'll need to make sure we are serving public assets:
  * `config.public_file_server.enabled = ESSI.config[:rails][:serve_static_files] || ENV['RAILS_SERVE_STATIC_FILES'].present?`
* The banner image has a resolution of 1800x400, rather than the Hyrax-recommended resolution of 5000x1000
  * We could resize it, but the closest we can get without distortion or cropping is 5004x1112
* This background color reveals "Search Digital Collections" next to the search bar, whose aesthetic we probably want to tweak


config/initializers/hyrax.rb
```diff
@@ -197,7 +197,7 @@
   # config.audit_user_key = 'audituser@example.com'
   #
   # The banner image. Should be 5000px wide by 1000px tall
-  # config.banner_image = 'https://cloud.githubusercontent.com/assets/92044/18370978/88ecac20-75f6-11e6-8399-6536640ef695.jpg'
+  config.banner_image = 'images/homepage_banner.png'
```
randalldfloyd, line 200: Can we check for a config variable here in case we want to switch it on the fly without pushing commits?  We don't even need to commit the corresponding key into the config YAML if we do a safe "dig" on the lookup, like:

`ESSI.config.dig(:essi, :homepage_banner) || 'images/homepage_banner.png'`

That way I could change the banner with a change to the secrets file and a restart.

randalldfloyd (changes requested):

config/initializers/hyrax.rb
aploshay: @randalldfloyd changes made, force-pushed

randalldfloyd (approved):

PR #287, opened 2021-04-03
At work creation, keep existing ocr_state value if present

aploshay:
Currently the config flag for `index_ocr_files` hard-sets ocr_state to "searchable", even if it's already been set to another value (though the manual creation form, or Bulkrax import parsing of CSV/METS/etc.)  This changes it to a soft set, only if blank.


randalldfloyd (approved):
@aploshay This is a great fix to allow import sources to dictate the value. Nick will likely write a follow-on story that looks into whether work types can selectively override and dictate the value.

PR #288, opened 2021-04-08
Adding okcomputer and custom service checks

randalldfloyd:
Adds the okcomputer gem and custom repository service checks.  **Note:** For this to be available in the application, there must be a configuration section in the main YAML config file.  It is being omitted from the committed config files so that a route for okcomputer is not defined by default in any given deployment. Configuration should look something like:
```
okcomputer:
    mount_at: '/status'
    user: user
    password: password
```


dlpierce (approved):
Looks OK ;)
Cantaloupe has a /health endpoint, but it's disabled currently on our services. Might be something to explore.

PR #292, opened 2021-04-16
[ESSI-1288] In AllinsonFlex, ensure empty properties return a Hash

aploshay:
When a work class was configured to use AllinsonFlex with no properties configured, the resultant `DynamicSchema` object had a `nil` value, instead of an empty Hash, in the `properties` field of the `schema` attribute (serialized JSON).  Granted, this is an odd situation you probably wouldn't _want_ to create -- but other parts of the stack assume a valid Hash value is present there, and break.  This includes the issue prompting the ESSI-1288 story, the form to create new work not rendering.  This monkeypatches that issue.

We may not actually want to merge this PR, as we can also simply avoid the situation until such time as it is fixed in AllinsonFlex.

### DIFF
--- a/lib/extensions/allinson_flex/dynamic_schema/load_empty_properties.rb
+++ b/lib/extensions/allinson_flex/dynamic_schema/load_empty_properties.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Extensions
+  module AllinsonFlex
+    module DynamicSchema
+      module LoadEmptyProperties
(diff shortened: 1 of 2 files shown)


aploshay:
Raised AllinsonFlex issue here:
https://github.com/samvera-labs/allinson_flex/issues/56

aploshay:
Closing as we can just avoid this, and it can be fixed directly in AllinsonFlex later.

PR #296, opened 2021-05-17
ESSI-1316 Making additional SMTP options configurable

randalldfloyd:
To get email to route out of container environments and to a SMTP server, different environments may need more control over the available Action Mailer options.

### DIFF
--- a/config/essi_config.example.yml
+++ b/config/essi_config.example.yml
@@ -31,6 +31,9 @@ default: &default
       user_name: 'example@gmail.com'
       password: 'app_password'
       authentication: 'plain'
+      raise_delivery_errors: false
+      enable_starttls_auto: true
+      openssl_verify_mode: 'none'
     serve_static_files: true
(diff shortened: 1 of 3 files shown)


config/essi_config.example.yml
```diff
@@ -31,6 +31,9 @@ default: &default
       user_name: 'example@gmail.com'
       password: 'app_password'
       authentication: 'plain'
+      raise_delivery_errors: 'false'
+      enable_starttls_auto: 'true'
```
aploshay, line 35: @randalldfloyd These are probably functionally equivalent but changing them from strings to raw true/false values would be more consistent with other config values in this section

randalldfloyd: @aploshay Yeah didn't even catch that I had done that.

aploshay (approved):

aploshay (approved):

PR #300, opened 2021-05-24
[ESSI-1157] Add work type selection to Bulkrax import form for METS

aploshay:
For METS format, changes Bulkrax format to pull work type from user form, rather than from METS metadata.

Another solution might be to handle this as rights_statement is done -- pulling the value from the metadata, but defaulting to the form value when blank, with a checkbox option to enforce overriding with the form value -- but this is minimal change necessary.


aploshay:
Retroactively converting to draft pending resolution of failing tests

aploshay:
Converted from draft to ready for review.  Tests are passing locally; CircleCI test failures last build were the `SQLite3::BusyException` sort.

aploshay:
Converting back to draft as I think this can be refactored into a more elegant solution

dlpierce (approved):

PR #308, opened 2021-06-04
[ESSI-1338] Log profile import separately, and provide in UI

aploshay:
Of note:
* This specifies a specific logfile for profile import, but I'm not familiar with how logging may occur differently in the containerized production environment -- does this need to be handled differently?
  * A related question is whether this file location should be specified in a config setting somewhere
* Adding in the link to log action in the profile import error flash message might be done in a more orthodox manner through a url/path helper


dlpierce:
In the container the Rails log is output to stdout, but writing to a separate log as done here should still work with the caveat that it would not be retained between service restarts. Making the log location configurable would be good as that would allow it to reside on a different storage volume.

lib/extensions/allinson_flex/prepend_importer.rb
```diff
@@ -0,0 +1,16 @@
...
+module Extensions
+  module AllinsonFlex
+    module PrependImporter
+      # modified from allinson_flex to use a separate log
+      def validate!
+        validator.validate(data: data, schema: schema, logger: validation_logger)
+      end
+
+      def validation_logger
+        @validation_logger ||= Logger.new(Rails.root.join('log', 'profile_validation.log'), 10, 2.megabytes)
```
dlpierce, line 12: Add log location configuration

dlpierce (changes requested):

lib/extensions/allinson_flex/prepend_importer.rb
aploshay: @dlpierce configuration added

dlpierce (approved):

PR #317, opened 2021-06-18
[ESSI-1361] Allow choosing Collection thumbnail from any Fileset (beyond first 10)

aploshay:
Inherited issues:
  * default limit of 10 rows in a solr query result (unless otherwise specified)
  * `SearchBuilder` and subclasses limit of 100 rows in a solr query result (even if a greater amount specified)
  * poorly-scaling performance of `#all_files_with_access` generating a tree of presenters and solr queries

Changes:
 * adds support for `CollectionMemberSearchBuilder` returning any configured number of rows
 * refactors #all_files_with_access to run a single solr query
 * adds indexing of `FileSet` titles to a work, along with ids
   *  This is more consistent with how collections associated to a work are indexed on the work for both titles and ids, and facilitates the `#all_files_with_access` refactoring

### DIFF
--- a/app/indexers/archival_material_indexer.rb
+++ b/app/indexers/archival_material_indexer.rb
@@ -4,6 +4,7 @@ class ArchivalMaterialIndexer < Hyrax::WorkIndexer
   include ESSI::IndexesArchivalMaterialMetadata # Replaces IndexesBasicMetadata
   include ESSI::ArchivalMaterialIndexerBehavior
   include ESSI::IIIFThumbnailBehavior
+  include ESSI::IndexesFilesets
   include ESSI::IndexesNumPages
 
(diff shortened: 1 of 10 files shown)


dlpierce (approved):
It seems like the ordering of the file_sets_ssim matches matches that of member_ids_ssim in a given instance.

aploshay:
> It seems like the ordering of the file_sets_ssim matches matches that of member_ids_ssim in a given instance.
@dlpierce that's a good point, this relies on the ordering matching -- if it turns out not to be reliable, the quickest available fallback would be to use the member_id value for both the display and value in the select field.  It's slightly less user-friendly, but guaranteed to match, and also guaranteed to be unique.

PR #320, opened 2021-06-30
[ESSI-1359] dependency support for IU Login upgrade

aploshay:
Minimally upgrades gems, yarn packages to support newer puma.  Holds back from newer versions of browse-everything, Hyrax, Rails to avoid resultant issues with webpacker and others.  Upgrades docker build and moves circleCI from sqlite to mysql database to avoid locking errors on feature specs.


dlpierce:
Getting the docker compose stack updated to use SSL presented a chance to get a bunch of outstanding config changes and docker image updates that I've been testing locally for a while merged.

PR #322, opened 2021-07-06
[ESSI-1322] Hide certain fields from non-admin display via blacklight

aploshay:
Changes here to later be reverted by ESSI-1321, which will make non-admin property visibility configurable instead of hard-coded.


dlpierce (approved):
Looks good as a quick fix.

PR #327, opened 2021-07-29
Essi-1329 Speedup work edit forms

dlpierce:
Loading the edit form for works with many filesets can be very slow, to the point of the web proxy timing out. Rendering the edit form loads from fedora all members of the work, both filesets and works, and then filters to only works. This PR changes the loading behavior to use the membership data indexed in solr.

The work file manager shows icons for each file indicating if there is an associated OCR file. This required a fedora lookup for each file. This is being changed to use the ocr_text solr field.

This also adds some profiling gems to the development gem group. Now, when in development mode, we should see a profiler dialog in the upper left of the screen. A stack profile for a particular page can be accessed by adding `pp=flamegraph` to the page url parameters.

### DIFF
--- a/app/helpers/file_manager_helper.rb
+++ b/app/helpers/file_manager_helper.rb
@@ -1,6 +1,6 @@
 module FileManagerHelper
   def ocr_check(presenter)
-    return '✅' if presenter.extracted_text
+    return '✅' if presenter.extracted_text?
     '❌'
   end
 end
--- a/app/helpers/file_manager_helper.rb
+++ b/app/helpers/file_manager_helper.rb
@@ -1,6 +1,6 @@
 module FileManagerHelper
-  def ocr_check(id)
-    return '✅' if FileSet.find(id).extracted_text
+  def ocr_check(presenter)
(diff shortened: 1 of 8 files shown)


app/helpers/file_manager_helper.rb
```diff
@@ -1,6 +1,6 @@
 module FileManagerHelper
-  def ocr_check(id)
-    return '✅' if FileSet.find(id).extracted_text
+  def ocr_check(presenter)
+    return '✅' if presenter.extracted_text
```
aploshay, line 3: Seems like we probably want to use the `extracted_text?` method here, or drop the method entirely

aploshay (approved):
I submitted an inline comment about possibly changing the `extracted_text` call to `extracted_text?`, but everything looks great and runs great so I'm also approving the PR as-is

andjsmit (approved):
Did a desk check with several text rich pages. OCR was generated and IIF search box worked. Found OCR in SOLR as configured. Search was quick.

PR #332, opened 2021-08-23
[ESSI-1371] In bulkrax export, skip redundant record saves, and make optional

aploshay:
Bulkrax export saves each record it exports, setting the `source` field to include the object's `id` value, to support round-tripping.  This slows down the export process results in server response timeout.

The first two commits add a check to only perform the save when the `source` does not yet contain the desired value.  This would improve the performance on subsequent exports of the same records, skipping unnecessary `save` calls.

The later commits add a new flag to the exporter object, form, and controller to allow opting out of the save calls, if the ability to round-trip is not needed.


dlpierce (changes requested):
There are a large amount of changes in the schema file aside from the intended additional column. These probably come from using sqlite vs mysql. I'll see if running db:migrate from in docker will clean this up.

dlpierce (approved):

PR #335, opened 2021-08-31
Essi-1323 Sequential job enqueueing

dlpierce:
Attempts to change the behavior of all jobs based on ApplicationJob to wait to enqueue new jobs until the current job has completed performing. This uses thread local variables to store the new jobs until an after hook completes the enqueueing.

Doing this helps avoid race conditions where the repository objects in memory in the primary import job thread do not match what has been written to active fedora by secondary jobs (e.g characterization, derivatives, etc).

Thanks to `Hyrax::ApplicationJob`, all jobs should now make use of this behavior including those contained within the hyrax gem.

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -10,12 +10,12 @@ class ApplicationJob < ActiveJob::Base
   end
   
   after_perform do |job|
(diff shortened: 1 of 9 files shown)


randalldfloyd (approved):
This looks solid.  It passes specs and desk-checks locally for me.

PR #339, opened 2021-09-07
essi-1441 Disable FITS use of tesseract via apache tika

dlpierce:
By default FITS uses Apache Tika as one of its tools, which in turn will run tesseract. As far as I can tell, no data from running tesseract in this way makes it into the repository. Other Samvera projects have disabled Tika in the same way as this PR, by commenting the relevant line out in the FITS configuration file. Our own tesseract OCR service should be unaffected.


aploshay (approved):
+1 by inspection

PR #356, opened 2021-10-29
ESSI-1323 Prevent some FileSet save race conditions

dlpierce:
Prevent race conditions causing missing FileSet metadata.


aploshay (approved):
+1 by inspection

randalldfloyd (approved):
+1 Deferring to @dlpierce on this one, especially given his intense deep-diving into all things job/race conditions.

PR #362, opened 2021-11-18
essi-1518 Remove shx dependency to fix asset precompilation

dlpierce:
Remove shx dependency to fix asset precompilation. Same change is in hyrax 3.


aploshay (approved):
+1 by inspection

PR #364, opened 2021-11-30
[ESSI-1370] bulkrax export display

aploshay:
It looks like in our change from bulkrax 1.0.0 to 1.0.2 a few months ago, there was a change to how 'collections' is parsed and stored that broke a view.  This attempts to work around the issue by differentially displaying the value.

Display is broken for both exporters currently up on the dev server.  Locally, I am finding display of some exporters works, others are broken, but this change resolves them all.  To test this locally, you may need to create an exporter if you currently have none.


wang58:
The fix looks good. The export displays well for all export from resources (collections, work type or importer)

PR #365, opened 2021-11-30
[ESSI-1519] collection lookup performance

aploshay:
Populating the Collections search auto-suggest involves building a CatalogController just to pull the repository value from it.  This entails loading allinson_flex and all the ProfileText lookups, even though we don't need them.

This PR addresses that performance hit by conditionally skipping loading allinson_flex in the CatalogController.  (But there might be a more elegant way of getting `repository` from another source?)

Testing this locally: you can start a Collection lookup with browser inspector open, and watch network activity -- it'll be calling some URL like:
https://localhost:3000/authorities/search/collections?access=deposit&q=SOME_SEARCH_TERM&_=SOME_NUMBER
You can open that in a new tab, and watch server activity while you load result.  Without this PR, you should see a lot of AllinsonFlex activity; with it, that should go away, and the response should be faster.


wang58:
This fix looks good to me. After the fix, I saw the performance improved and the logs bypass the unnecessary allinson_flex load.

PR #369, opened 2021-12-28
[ESSI-1539] tweak dynamic indexing

aploshay:
The default indexing process returns an `Array` for multi-valued field values, in a `#to_solr` call:
https://github.com/samvera/active_fedora/blob/12.2-stable/lib/active_fedora/indexing_service.rb#L30-L65
https://github.com/samvera/active_fedora/blob/12.2-stable/lib/active_fedora/rdf/indexing_service.rb#L21-L83

However, the dynamic indexing from `allinson_flex` pulls values directly from the resource:
https://github.com/IU-Libraries-Joint-Development/allinson_flex/blob/f3ed44ff96d50bc7c42cf2b9693d9b081f9a2f0c/app/indexers/allinson_flex/dynamic_indexer_behavior.rb#L20
which results in a value type of
```rb
ActiveTriples::Relation
```
instead.

If this is saved to `Solr`, it is ultimately persisted as an `Array` -- and that's what we get back from a `Solr` search, and all is well.  But if you directly build a `SolrDocument` from calling `#to_solr` on a Work, you'll get test-breaking results like:
```rb
> solr_doc.title
=> [['Some Title']] # an Array wrapping an ActiveTriple::Relation, instead of a simple Array
```
This PR resolves that by adding an another dynamic indexing module, that fixes datatypes at the end of the `#generate_solr_document` stack.

Keeping in draft status for now, for a few reasons:
* we should merge #368 first
* we may be okay leaving dynamic indexing results as they are (closing this PR without merging)
* we may prefer to resolve with issue directly within allinson_flex, or through a different resolution within essi

### DIFF
--- a/app/indexers/archival_material_indexer.rb
+++ b/app/indexers/archival_material_indexer.rb
@@ -14,6 +14,7 @@ class ArchivalMaterialIndexer < Hyrax::WorkIndexer
   #  end
   # end
   include AllinsonFlex::DynamicIndexerBehavior
+  include ESSI::DynamicIndexerBehavior
   self.model_class = ::ArchivalMaterial
 
 end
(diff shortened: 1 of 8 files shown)


dlpierce (approved):
This looks good. Easy to remove if it is later addressed within allinson flex.

PR #370, opened 2021-12-29
[ESSI-1542] Fix CircleCI builds with ruby holdback

aploshay:
Upgrading from ruby 2.7.4 to 2.7.5 should not, in principle, break the stack.  But, it does.

In the future, this might be resolved in updates to the affected gems, or with ruby 2.7.6.

But for the moment, holding back to 2.7.4 keeps the build working.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,7 @@ jobs:
   build:
     docker:
     # legacy needed for phantomjs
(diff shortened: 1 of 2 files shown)


dlpierce (approved):
We do need to resolve this at some point. I've not seen any reports of encountering this issue in other software, so it might be something odd we (or the oai or bulkrax gems) are doing.

PR #372, opened 2022-01-03
Essi 1507 collection source identifier

aploshay:
In draft status pending:
* setting a proper predicate for the source_identifier property
* possibly changing source/source_identifier mapping config for CSV, METS import


rwingiupui:
+1 Desk Checked - Source identifier field visible and available when creating a collection.

aploshay:
Reverting to draft status pending some more discussion with POs over metadata handling

aploshay:
Closing without merging, due to decision to continue using "source" field for bulkrax import/export

PR #373, opened 2022-01-03
[ESSI-1540] Bugfix and refactoring for allowing PDF download

aploshay:
Various changes:
* Moves shared Presenter code into a module, and shared Presenter spec code into a shared_example
* Changes the paradigm for the pdf download Boolean to match the pattern we're using for ocr_state/ocr_searchable
  * replaces allow_pdf_download with pdf_state (ActiveFedora property) and pdf_downloadable (solr index)
    * pdf_state is present in the minimal default metadata profile, but the property is not hard-coded, rather expected to be configured via allinson_flex
  * adds a custom UI component to present pdf_state values ("downloadable", "disabled") as a true/false radio button
  
For this to work, there needs to be a pdf_state property configured in allinson_flex, as follows:
**key**: pdf_state
**display_label**: (probably "PDF Download" or something similar)
**available_on**: Paged Resource and ArchivalMaterial, minimally -- could apply to other work types, too, but we'll need to start including the PDF download modules for them to actually have pdf download work
**cardinality**: min 0, max 1
**indexing**: stored_searchable, and (probably) admin_only
**usage_guidelines**: might want something here about how admin status overrides a "false" here?
**property_uri**: suggest "http://dlib.indiana.edu/vocabulary/PDFState" to mimic what we have for ocr_state, but the code just cares about the property name (pdf_state)

### DIFF
--- a/app/models/concerns/essi/pdf_behavior.rb
+++ b/app/models/concerns/essi/pdf_behavior.rb
@@ -3,8 +3,7 @@ module ESSI
     extend ActiveSupport::Concern
 
     def pdf_downloadable?
-      return true if self.pdf_state == 'downloadable'
-      return false
+      self.pdf_state == 'downloadable'
     end
 
     def to_solr(solr_doc = {})
--- a/app/models/concerns/essi/pdf_behavior.rb
+++ b/app/models/concerns/essi/pdf_behavior.rb
@@ -1,0 +1,17 @@
+module ESSI
+  module PDFBehavior
+    extend ActiveSupport::Concern
+
+    def pdf_downloadable?
+      return true if self.pdf_state == 'downloadable'
+      return false
+    end
+
+    def to_solr(solr_doc = {})
+      super.tap do |doc|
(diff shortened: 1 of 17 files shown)


app/models/concerns/essi/pdf_behavior.rb
```diff
@@ -0,0 +1,17 @@
+module ESSI
+  module PDFBehavior
+    extend ActiveSupport::Concern
+
+    def pdf_downloadable?
+      return true if self.pdf_state == 'downloadable'
+      return false
```
dlpierce, line 7:
```suggestion
      self.pdf_state == 'downloadable'
```

dlpierce (changes requested):

app/models/concerns/essi/pdf_behavior.rb
aploshay: @dlpierce resolved in added commit

dlpierce (approved):

PR #379, opened 2022-01-18
Optimize profile property text loading to avoid N+1 db queries

dlpierce:
Eager loads profile properties and their texts to avoid querying the db thousands of times.


aploshay (approved):
Change looks good, tests well in console, and the involved database tables already have the pertinent fields indexed

PR #380, opened 2022-01-19
Cache data from holding location service

dlpierce:
The holding location web service is being queried for every work on a search results page, which adds a second or two to the response time when all 10 results have a holding location.

This caches the web service response for between 1-2 hours.


aploshay (approved):
+1 by inspection

PR #381, opened 2022-01-21
ESSI-1554 Load AllinsonFlex attributes only when needed

dlpierce:
The AllinsonFlex DynamicSolrDocument is loading the profile and (re)defining attributes every time a SolrDocument instance is initialized. On pages with multiple objects this incurs a large performance penalty. For example, the All Collections Dashboard test case would spend ~80% of its time in this method.

This loading only needs to happen once per request, but because the outcome is defining methods, the effect of loading should persist even after the request is finished.

### DIFF
--- a/app/models/concerns/essi/dynamic_solr_document.rb
+++ b/app/models/concerns/essi/dynamic_solr_document.rb
@@ -17,13 +17,21 @@ module ESSI
         # profile directly.
         profile = AllinsonFlex::Profile.current_version
         unless profile.blank?
-          profile.properties.each do |prop|
-            attribute(
-              prop.name,
-              # if the property is singular, make it so
-              prop.cardinality_maximum == 1 ? Hyrax::SolrDocument::Metadata::Solr::String : Hyrax::SolrDocument::Metadata::Solr::Array,
-              solr_name(prop.name.to_s)
-            )
+          Rails.logger.debug { "Loading AllinsonFlex DynamicSolrDocument #{ profile.profile_version } with #{profile.properties.count} properties." }
+
+          # Loading attributes is done by defining a method for each property. This only needs to be done once or when the profile changes.
+          # Code that needs to run for every SolrDocument instance should go outside this block.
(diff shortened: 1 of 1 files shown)


app/models/concerns/essi/dynamic_solr_document.rb
```diff
@@ -17,13 +17,21 @@ def load_allinson_flex
         # profile directly.
         profile = AllinsonFlex::Profile.current_version
```
aploshay, line 18: Would it be beneficial to eager load .properties here?

dlpierce: We could, but this code doesn't load property associations in the way DynamicCatalogBehavior does. It should only ever need 3 sql queries at most and that includes the count in the log string.

aploshay (approved):

app/models/concerns/essi/dynamic_solr_document.rb
```diff
@@ -17,13 +17,21 @@ def load_allinson_flex
...
-          profile.properties.each do |prop|
-            attribute(
-              prop.name,
-              # if the property is singular, make it so
-              prop.cardinality_maximum == 1 ? Hyrax::SolrDocument::Metadata::Solr::String : Hyrax::SolrDocument::Metadata::Solr::Array,
-              solr_name(prop.name.to_s)
-            )
+          Rails.logger.debug { "Loading AllinsonFlex DynamicSolrDocument #{ profile.profile_version } with #{profile.properties.count} properties." }
+
+          # Loading attributes is done by defining a method for each property. This only needs to be done once or when the profile changes.
```
orangewolf, line 22: @dlpierce I know this just got merged, but once you feel like its battle tested, I think this would be something that Allinsonflex should just have by default. do you agree?

dlpierce: Yes, certainly. There are a few enhancements we've come up with that we are planning to move into the gem.

PR #382, opened 2022-01-21
Cache user ldap roles to speed up user.admin?

dlpierce:
Calling `admin?` on a user currently regenerates the user's ldap membership list. Some metadata fields now check admin status causing repeated lookups to render a single page. This change caches that result. The cache entry will be invalidated by a change to the user record, which should happen when logging in.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -45,8 +45,11 @@ class User < ApplicationRecord
 
   # Roles to add depending on user's LDAP groups and ESSI configuration
   def ldap_roles
-    mappings = ESSI.config.dig(:ldap, :group_roles) || {}
-    mappings.select { |role, groups| member_of_ldap_group?(groups) }.keys
+    Rails.cache.fetch("ldap_roles-v1-#{cache_key_with_version}",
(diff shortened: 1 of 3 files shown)


aploshay (approved):

aploshay:
@dlpierce  Added a commit with some changes, on my personal fork.  Don't know if you necessarily want to incorporate any/all of them, but they do get the broken test passing:
https://github.com/aploshay/essi/commit/f10989d02c17c5c6eae5d81de55b9557fb11b318

dlpierce:
Looks great, but I don't see why it fixes specs on circle. I haven't seen any broken specs locally, so it's been tough to diagnose.

aploshay:
Spooky!  Running 2a42464, I get the same spec failure locally that CircleCI does, and that you don't.  Running bf1c65c, the specs pass for me locally, but fail on CircleCI only for the cache wrapper method.

dlpierce:
Running just users_spec.rb does not fail, but running all specs does fail for me. It seems like it is a cache issue, but I don't know why the commit that disabled cache still failed.

aploshay (approved):

PR #388, opened 2022-02-03
[ESSI-1562] Fix catalog "masonry" and "gallery" views with query parameter, add to "list" view

aploshay:
An earlier PR added functionality to pass along a search term through catalog results, to the individual work display, to be applied in UV for search term highlighting:
https://github.com/IU-Libraries-Joint-Development/essi/pull/90

It looks like there's been some gem drift in the meantime that broke our monkeypatch.  The PR fixes the monkeypatch, which should resolve the ruby errors that can currently be generated in "gallery" and "masonry" view for catalog results that include Collections.

The second commit adds the text query pass-through feature to the catalog results "list" view, where it was missing.


randalldfloyd (approved):
Seems correct on inspection.

PR #391, opened 2022-02-14
Essi-1545 Browse Everything optimized

dlpierce:
Optimizes how Browse Everything uses the DOM to track which files are selected.

Selecting 12000 files can now be done in under a minute, from opening of dialog to submitting the form, where previously it took ~10 minutes just to select the files in a directory.

Once we update the browse-everything gem to a version that has this fix (> 1.1.2?), this file should be removed.


aploshay (approved):
+1 by inspection (and fully-passing test suite!)

PR #394, opened 2022-02-18
ESSI-1576 Quiet route errors

dlpierce:
Going to a missing page or unknown format is usually just a 404 sent to the client. Raising an exception and airbrake notice seems unnecessary.

Based on the new CircleCI config branch.


aploshay (approved):
Desk-checks!

PR #395, opened 2022-02-18
ESSI-1577 Sidekiq dashboard link

dlpierce:
Add a Sidekiq dashboard link to the hyrax dashboard. Only visible to admins, which is now checked via the user's abilities to allow other user types to potentially access it in the future, if needed.

Note to devs: I could not find a route helper (e.g. sidekiq_web.root_path) so had to use a string instead.

Based on the new CircleCI config branch.


aploshay (approved):
Desk-checked

PR #401, opened 2022-03-03
Configurable switch_user in development without login

aploshay:
If configured to "true" for local development, allows direct access to switch_user without first logging in.

(_Not_ tagging this for configuration change, since we don't need or want any configuration change on the servers.)

### DIFF
--- a/app/views/_user_util_links.html.erb
+++ b/app/views/_user_util_links.html.erb
@@ -25,13 +25,28 @@
         <li><%= link_to 'Log In As ...', main_app.log_in_as_path %></li>
         <% end %>
       </ul>
-
     </li><!-- /.btn-group -->
   <% else %>
     <li>
       <%= link_to main_app.new_user_session_path do %>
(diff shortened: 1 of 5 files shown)


andjsmit:
Ran into 2 issues during testing.

1. Prior to initial login, click the "remember user" check box produces error (as there is no user to remember)
2. Not possible to select first user on initial login.

Neither of these are necessarily deal breakers as this is for development use only.

andjsmit (changes requested):
Can https://github.com/IU-Libraries-Joint-Development/essi/pull/401/files#diff-0b4214521bfb49c7ee5b524db9d5796e0298022a9ab59040bd42a1f9b7871e17L51 be expanded as well?

```
config.switch_back = (Rails.env.development? && ESSI.config.dig(:essi, :allow_dev_login)) ? false : true
```

The option to switch back is not needed if switch user is enabled for everyone and in fact causes an error if clicked prior to be logged in.

aploshay:
@andjsmit pushed a couple more commits and rebased

PR #404, opened 2022-03-23
Catch unusable admin sets before work creation

aploshay:
An issue I sometimes have locally is that I'll try to create a work using an AdminSet that lacks an active_workflow (or a PermissionTemplate on which to hang an active_workflow), and get a Ruby error after submission.

This moves that check into the UI, before selecting an AdminSet.  It also changes the select population to use Solr instead of ActiveFedora for looking up AdminSet titles.


aploshay:
unsuable_admin_sets

PR #406, opened 2022-03-29
[ESSI-1566] Catalog sorting collections by "size"

aploshay:
The final commit isn't strictly necessary, as it's just copying that file in from hyrax without modification, but I felt inclined to make that hidden, inherited setting about the maximum nesting depth (5) explicit within the project.

In draft status pending adding specs.


randalldfloyd (approved):
I've been looking at this as part of ESSI-1590, so I'm pretty familiar with it.

PR #407, opened 2022-03-29
[ESSI-1595] Indexing fixes for catalog sorting

aploshay:
Stacks on top of #406 
Adds indexing for missing sortable fields (date_created, sort_title, number_of_pages) and tweaks indexing for num_works, num_collections.

Note that an object will have to be saved to adopt the updated indexing.  #406 provides a rake task to update collections, which may need to be regularly scheduled, but this PR doesn't include anything to do a one-time update of works.  I'm applying the "configuration change" label to capture this. 

Force-pushed an update: in essi-1566, forgot to add the "size" sorting link to view _all_ collections (along with collections by campus); added it.


randalldfloyd (approved):
This all makes sense to me, though I didn't have time to desk check it.

PR #409, opened 2022-03-30
ESSI-1560 add works number and colections number in the homepage

wang58:
Add the current time, the number of collections, and the number of works on the homepage.


rwingiupui:
+1 Desk checked - correctly increments/decrements number of work(s) and collection(s) - displays information on homepage.

PR #413, opened 2022-04-01
[ESSI-1582] Change configured text to show child images

aploshay:
@nhomenda can you confirm desired language?
Setting as Draft pending fixing specs


nhomenda:
@aploshay thanks, I like "Item Images"

aploshay (commented):
Self-approving

PR #416, opened 2022-04-05
[ESSI-1598] Restore hyrax based_near behavior allinson_flex overrides

aploshay:
For `based_near`:
* restores normal hyrax indexing behavior
* injects `class_name: Hyrax::ControlledVocabularies::Location` into properties for the property

Without these, configuring a `based_near` property in allinson_flex breaks indexing of a work persisting any values to the attribute, raising a ruby error

### DIFF
--- a/app/models/archival_material.rb
+++ b/app/models/archival_material.rb
@@ -30,5 +30,6 @@ class ArchivalMaterial < ActiveFedora::Base
   # schema (by adding accepts_nested_attributes)
 
   include AllinsonFlex::DynamicMetadataBehavior
+  include ESSI::DynamicMetadataBehavior
   include ::Hyrax::BasicMetadata
 end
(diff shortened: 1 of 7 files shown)


randalldfloyd (approved):
@aploshay This looks great, though I'm just curious what it is that Hyrax itself is being opinionated about with based_near.  I see it in a *lot* of upstream code.

aploshay:
@randalldfloyd there's `#append_to_solr_doc` that comes from ActiveFedora::
https://github.com/samvera/active_fedora/blob/12.2-stable/lib/active_fedora/rdf/indexing_service.rb#L42-L52
which hyrax overrides:
https://github.com/samvera/hyrax/blob/2.x-stable/app/indexers/hyrax/deep_indexing_service.rb#L4-L20
and ends up calling a `#solrize` method:
https://github.com/samvera/hyrax/blob/2.x-stable/app/indexers/hyrax/deep_indexing_service.rb#L63-L77
which does not exist on a standard `ActiveTriples::Resource` but which is provided by the `Location` class that `based_near` is expected to use:
https://github.com/samvera/hyrax/blob/2.x-stable/lib/hyrax/controlled_vocabularies/location.rb

PR #417, opened 2022-04-06
[ESSI-1570] collection save sets default thumbnail

aploshay:
Sets a default thumbnail, if none is currently set and a default value is available, after a Collection _update_.  (Lookup is not possible until the Collection has been persisted.)

### DIFF
--- a/spec/models/collection_spec.rb
+++ b/spec/models/collection_spec.rb
@@ -1,5 +1,95 @@
 require 'rails_helper'
 
 RSpec.describe Collection do
-  # "Add your tests here"
+  let(:collection_type) { double(:brandable, true) }
+  let(:collection_type_gid) { FactoryBot.create(:user_collection_type).gid }
+  let(:collection) { Collection.create(title: ['Persisted collection'],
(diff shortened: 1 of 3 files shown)


spec/models/collection_spec.rb
```diff
@@ -1,5 +1,82 @@
...
+
+      before do
+        allow_any_instance_of(Hyrax::Forms::CollectionForm).to \
+          receive(:send).with(:all_files_with_access).and_return([])
+      end
+      it "returns nil" do
+        expect(collection.default_thumbnail_id).to be_nil
+      end
+    end
+    context "with no thumbnails available" do
```
dlpierce, line 25:
```suggestion
    context "with thumbnails available" do
```

aploshay: @dlpierce re-pushed fixing this and with some refactoring, changing the before_save callback to a before_update

dlpierce (approved):

aploshay:
@dlpierce I think this is ready, I'm just reverting it to draft as a way to indicate we should hold off on this until we get an error-free 1.2.x

PR #422, opened 2022-04-15
Essi-1604 Increase presenters' solr row limit

dlpierce:
I noticed a paged resource with 1111 pages only rendered 1000 pages in UV. This fixes that by raising the limit hard coded in Hyrax::PresenterFactory. Tested with an adapted version of the hyrax presenter_factory_spec.rb


aploshay (approved):
+1 by inspection

PR #423, opened 2022-04-15
Essi-1606 Use solr terms query parser

dlpierce:
Hyrax::GrantEditToMembersJob was failing on works with a large amount of members due to exceeding solr's boolean clause limit. This switches to using the terms query parser to supply the list of ids.


lib/extensions/hyrax/jobs/file_set_ids_post.rb
```diff
@@ -3,8 +3,9 @@ module Hyrax
     module Jobs
       module FileSetIdsPost
         # POST to avoid URI Too Long error from solr, and raise row limit
+        # Uses solr terms query parser to support large (> 1024) amounts of member_ids
         def file_set_ids(work)
-          ::FileSet.search_with_conditions({id: work.member_ids},
+          ::FileSet.search_with_conditions(ActiveFedora::SolrQueryBuilder.construct_query_for_ids(work.member_ids),
```
aploshay, line 8: @dlpierce looks like changing `ActiveFedora` to `::ActiveFedora` will fix the failing specs.  I also prefer `10_000` to `10000` but that's just a style preference.

aploshay, line 8: @dlpierce looks like changing `ActiveFedora` to `::ActiveFedora` will fix the failing specs.  I also prefer `10_000` to `10000` but that's just a style preference.

dlpierce: Good call, thanks!

aploshay (approved):

PR #424, opened 2022-04-19
ESSI 1578 search within collections

randalldfloyd:
Currently, in both Hyrax and Digital Collections, searching within a collection does not produce the expected results when a matching work is known to exist.  The primary reason is that the search fields in the Solr query are not the same as those in the overall global search.  Once that is remedied, then the next thing that is apparent is that the search is constrained to just the immediate members and not anything in the tree for nested subcollections.  The expectation would be that works throughout the entire tree would be searched, not just the direct membership at any one level.

This PR patches the main Collections controller to add query fields for the first part.  For recursive searching of the collection tree, this adds a new search builder that can build an array of collection IDs to search, and overrides CollectionMemberService to choose it based on a query parameter in the request.


aploshay (approved):
+1 by inspection

PR #428, opened 2022-05-03
ESSI-1578 Adds Solr logic to collection search to find FileSets with matches in full text

randalldfloyd:
The previously merged PR #424 ensured recursive searching through the subcollection tree and to also make sure the same fields were being searched as those in the default global search.  But, it failed to consider that it's the FileSets that actually contain the searchable text from OCR extraction, and it only searched for works.

This PR adds to the search processing chain by calling a method in a Hyrax search builder that finds and joins matching child FileSet IDs.  To do so, it also changes which search builder to inherit from instead of the default Blacklight builder.


aploshay (approved):
Desk-checks

PR #430, opened 2022-05-06
Essi-1610 Eliminate fedora request during manifest generation

dlpierce:
Generation of manifests is significantly slowed down by making a fedora request for every fileset. Removing this check might allow a manifest to include a missing image if the fileset has disappeared from fedora for some reason, but otherwise should not have any detrimental effect.

Increased visibility of such a problem might even be benefical?

### DIFF
--- a/app/presenters/hyrax/displays_image.rb
+++ b/app/presenters/hyrax/displays_image.rb
@@ -12,7 +12,7 @@ module Hyrax
     #
     # @return [IIIFManifest::DisplayImage] the display image required by the manifest builder.
     def display_image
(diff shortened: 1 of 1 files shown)


aploshay:
We _could_ replace the fedora request with an equivalent solr request, but I'm all for dropping the check entirely in return for the performance boost

aploshay (approved):

dlpierce:
The `DisplaysImage` module only gets included in `FileSetPresenter` so it seems implied there is a solr record for the fileset.

PR #432, opened 2022-05-12
[ESSI-1482] Add action to (re)generate extracted_text OCR

aploshay:
* adds a new job to call `create_ocr_derivatives` instead of `create_derivatives` on the FileSet
  * currently, all `create_derivatives` does is call `create_ocr_derivatives`, but relying on that seemed brittle
* adds a new `regenerate_ocr` controller action to call the job
* adds a link to the action
  * for some reason the path helper method doesn't exist, so I kludged around it :/

### DIFF
--- a/app/jobs/create_ocr_job.rb
+++ b/app/jobs/create_ocr_job.rb
@@ -1,0 +1,18 @@
+class CreateOCRJob < CreateDerivativesJob
+  queue_as Hyrax.config.ingest_queue_name
+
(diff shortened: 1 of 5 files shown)


dlpierce:
Am I correct that this is only available to admin users since Ability is not modified?

dlpierce:
Just as long as it's not available to regular users it should be fine.

dlpierce (approved):

PR #435, opened 2022-05-20
[ESSI-1609] Improve ingest scaling

aploshay:
DRAFT PR

Current contents:
  * improve scaling of `from_url: true` case (applies to any work creation)
Possible additions:
 
 Possible additional content:
 * reconsider `file_set.parent.update_index if parent_needs_reindex?(file_set)` in `CreateDerivativesJob` 
 * reconsider duplicated calls beween `VisibilityCopyJob`, `InheritPermissionsJob`

### DIFF
--- a/app/jobs/inherit_permissions_job.rb
+++ b/app/jobs/inherit_permissions_job.rb
@@ -1,0 +1,25 @@
+# imported from hyrax
+# A job to apply work permissions to all contained files set
+#
+class InheritPermissionsJob < Hyrax::ApplicationJob
+  # Perform the copy from the work to the contained filesets
+  #
+  # @param work containing access level and filesets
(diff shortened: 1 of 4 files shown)


aploshay:
Local testing suggests that the hyrax code modified here does NOT actually lead to triangular growth as the number of FileSet objects increases.  Although in the S3 file upload case, #perform_later calls are made to VisibilityCopyJob and InheritPermissionsJob per FileSet -- instead of just once, after all files have been handled -- since they are always made with the same single argument (the parent work), the job only ends up getting run once, as duplicate #perform_later calls are apparently collapsed into a single run of the job.  This approach still rubs me the wrong way, but it doesn't seem to perform problematically in the way I feared.

PR #436, opened 2022-05-24
[ESSI-1545] Fix bulkrax import post-unzip file selection

aploshay:
Prior file selection logic was grabbing _every_ previously-imported XML file along with the current one.
This fixes that by copying the correct logic over from the current bulkrax `XmlParser`.


randalldfloyd (commented):
That looks correct; I won't get bogged down in trying to desk-check it.

PR #440, opened 2022-06-02
Essi-1612 Double the stack size limits to allow even larger works

dlpierce:
Should allow SaveStructureJob to succeed with works with 1300+ pages


aploshay (approved):
+1 by inspection

PR #443, opened 2022-06-09
Essi-1575 Run javascript tests using karma

dlpierce:
This PR changes our javascript jasmine specs to run using the Karma JS library. It removes the jasmine ruby gems, as they are no longer updated, instead installing jasmine via yarn. It also removes phantomjs, instead using chrome, either by way of webdrivers or a container in docker-compose, depending on the environment.

The spec:javascript rake task has been replaced with a spec file that execs karma. To run only javascript specs from inside the docker-compose essi container:
`$ bundle exec rspec spec/javascripts/jasmine_spec.rb`

Running locally outside docker will require the ENV vars KARMA_BROWSER and KARMA_HOSTNAME to be set.


aploshay (approved):
Desk checks (as long as assets are clobbered first, as discussed on slack)

PR #449, opened 2022-06-14
[ESSI-1616] "description" display configurable via allinson_flex

aploshay:
Property-specific view partials are used when available, instead of the default partial (which accounts for allinson_flex dynamic label and hint values).  Hyrax provides partials for some basic/core metadata, including "description", which prevents allinsonflex configuration for label, hint values taking effect.  The applied fix is to override the property-specific partial with a copy of the default partial.

The current list of hyrax-provided edit partials is available here:
https://github.com/samvera/hyrax/tree/v2.9.6/app/views/records/edit_fields
and the up-to-date list for the newest version of hyrax is always available here:
https://github.com/samvera/hyrax/tree/main/app/views/records/edit_fields

To review this change, you will need to have a "description" property configured via your local allinson_flex profile.

### DIFF
--- a/app/views/records/edit_fields/_description.html.erb
+++ b/app/views/records/edit_fields/_description.html.erb
@@ -1,6 +1,14 @@
-<%# imported from hyrax to override %>
+<%# modified from hyrax as modified copy of "default" partial %>
 <% if f.object.multiple? key %>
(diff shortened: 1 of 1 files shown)


randalldfloyd (approved):
Makes sense on inspection

PR #450, opened 2022-06-16
[ESSI-1556] Block blacklight bookmark actions that create persistent guest accounts

aploshay:
When called by an unathenticated user, bookmarks actions create one-off guest accounts in the database, that then stick around forever.  This change blocks those routes.


randalldfloyd (approved):
I like this simple approach vs. trying to undo underlying Devise and/or Blacklight guest handling behavior. It's nice to have those revealed to us due to the guest account problem because there may be even better reasons for blocking routes that expose POST actions but that have no purpose in the UI.

PR #452, opened 2022-07-01
Essi-1491 Fix duplicated event handlers

dlpierce:
Moving collections.js to the same relative location as in hyrax allows it to override that code as intended.


aploshay (approved):
+1 by inspection

PR #454, opened 2022-07-13
Essi-1657 Update to Rails 5.2.8.1 and remove debase, ruby-debug-ide gems

dlpierce:
Addresses CVE-2022-32224
Removes some debugging gems from the Gemfile that Rubymine installs custom versions of.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ end
 
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '~> 5.2.4'
+gem 'rails', '~> 5.2.8'
 # Use sqlite3 as the database for Active Record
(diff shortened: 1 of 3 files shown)


rwingiupui:
+1 Desk checked - No issues found.

dlpierce:
Loading the profile dashboard still causes an error.

aploshay:
I have [a proposed change up on my fork](https://github.com/aploshay/essi/commit/9738586746a979eb0994cea9497060679cfbcbf6)

dlpierce:
I was going to try to adapt the solution blacklight used (SearchParamsYamlCoder), but I'm not clear on the need for the additional handling it does around yaml loading. If just modifying the AR setting works then I'm fine with that.

PR #455, opened 2022-07-14
[ESSI-1614] catch errors rendering #new with a new AllinsonFlex profile

aploshay:
There's a time window after a new AllinsonFlex profile has been saved where trying to render #new for a work generates a Ruby error.  This occurs when #load_allinson_flex in DynamicMetadataBehavior assumes dynamic_schema has returned a non-nil value, when that's the condition we're hitting.

The fix applied here:
* Adds a NoAllinsonFlexSchemaError case to parallel the NoAllinsonFlexContextError case
* in work models, catches and logs the errors
* in work controllers, catches and redirects the errors

### DIFF
--- a/app/controllers/concerns/essi/works_controller_behavior.rb
+++ b/app/controllers/concerns/essi/works_controller_behavior.rb
@@ -38,6 +38,15 @@ module ESSI
         ::IIIFManifest::ManifestFactory.new(presenter).to_h.to_json
       end
     end
+
+    # Overrides stock Hyrax method to catch AllinsonFlex errors during build_form
+    def new
+      begin 
+        super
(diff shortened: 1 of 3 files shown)


dlpierce:
Is there potential for bad side effects (such as a work saving incorrectly) if this error is not raised in the model?

aploshay:
> Is there potential for bad side effects (such as a work saving incorrectly) if this error is not raised in the model?

@dlpierce There shouldn't be any _new_ potential for bad side effects -- previously, a `load_allinson_flex` call just failed silently.  Now, if it fails, it's logged.

Works that have a `dynamic_schema_id` should all be fine, and only new works should be missing one.

PR #460, opened 2022-08-02
[ESSI-1646] improve flexible metadata reliability in bulkrax import

aploshay:
Awhile back, #279 attempted to address the issue of bulkrax import not having loaded flexible metadata properties, but ESSI-1646 captures the finding that this fix only works _sometimes_.  This should be a more robust fix.

There's a little bit of a chicken-and-egg problem here:
* we need to know the imported work's class, in order to call `factory_class.new` and load the most recent allinsonflex profile
  * the imported work's class is stored in `parsed_metadata` during `build_metadata`
    * `build_metadata` filters which properties are stored in `parsed_metadata` based on what it believes the work class can receive...
      * ...which is determined by calling `factory_class`

So, we need to call `factory_class` first in order to get`build_metadata` to build all the flexible metadata, but we need to call at least _some_ of `build_metadata` first to ensure that we get the desired `factory_class` (and not just defaulting to `PagedResource`).

The approach applied here is to call a new method, `add_work_type`, which serves as a sort of stripped-down version of `build_metadata`, before calling `factory_class`.

### DIFF
--- a/app/models/bulkrax/mets_xml_entry.rb
+++ b/app/models/bulkrax/mets_xml_entry.rb
@@ -60,13 +60,18 @@ module Bulkrax
       @files ||= record.files
     end
 
+    def add_work_type
+      self.parsed_metadata ||= {}
+      self.parsed_metadata['work_type'] = [parser.parser_fields['work_type'] || 'PagedResource']
+    end
+
     def build_metadata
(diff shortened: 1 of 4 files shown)


randalldfloyd (approved):
+1 on inspection only, best test will probably be in server deployment after merge.

PR #461, opened 2022-08-04
Essi 1488 purl 404

randalldfloyd:
The current default behavior of the PURL resolver when it fails to find a match is to redirect out to the old Variations access CGI.  This behavior came from Pages Online, which was predominantly musical scores.  What we want now is to redirect to the global 404 response within the application.  

We already had the ability to configure the rescue URL in the application configuration YAML, so this just changes the default to be a generic works endpoint that should always force the 404 response, like:

https://localhost:3000/concern/works/missingWorkID

Note the "configuration change" label as a reminder to set the value on deployed instances.  This path can be taken from the example config (`/concern/works/%s`), but could be anything that will be handled internally as a 404, like:

/404
/404.html
/404/%s (appends given ID)
etc.

This also restyles the `public/404.html` page to look like a modern 404 response instead of what looks like a generic site error. The style was approved by PO's:

DC_404-screenshot-2

### DIFF
--- a/config/essi_config.docker.yml
+++ b/config/essi_config.docker.yml
@@ -103,7 +103,7 @@ default: &default
     store_files_as_jp2: true
     kdu_compress_path: 'opj_compress' # set as kdu_compress or opj_compress, or '' to use ImageMagick
     master_file_service_url: http://purl.dlib.indiana.edu/iudl/variations/master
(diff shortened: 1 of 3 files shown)


aploshay (approved):
+1 by inspection

PR #462, opened 2022-08-05
[ESSI-1645] import collection field (via bulkrax METS import)

aploshay:
Adds optional selection of a Collection to use for METS ingest.

Note that unlike CSV ingest, which looks up a collection by title/source value and will create a collection if a match is not found, this only finds an existing collection that the user has deposit rights to.  See wiki page "Bulkrax: collection import and association" for more details.


randalldfloyd (approved):
+1 on inspection, haven't actually tried it out locally.

PR #465, opened 2022-08-11
[ESSI-1665] add CapybaraWatcher gem; add wait call in csv import spec

aploshay:
The theory on the csv import spec non-deterministically failing is that it's hitting a race condition on the closing the modal and the updating of the rest of the page.

This attempts to address that by adding the CapybaraWatcher gem, and adding a wait call around checking for "Cloud Files Added"

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,6 +43,7 @@ group :development, :test do
   # Adds support for Capybara system testing and selenium driver
   gem 'capybara'
   gem 'capybara-screenshot', '~> 1.0'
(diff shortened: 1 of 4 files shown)


aploshay:
@randalldfloyd It's got 5 successful runs in CircleCI, so it seems stable now

randalldfloyd (approved):
+1 Good find on the capybara_watcher gem.  The repo for it looks like it might be at risk of not being kept current, but the code is really simple.  We could fork it and take care of it if it ever came to that.

PR #470, opened 2022-08-16
[ESSI-1093] Add "Terms of Use" field to collections using rights_note property.

andjsmit:


app/views/records/show_fields/_rights_note.html.erb
```diff
@@ -0,0 +1 @@
+<span itemprop="rights_note" class="detail-width-250"><%= record.solr_document.rights_note[0] %></span><br />
```
aploshay, line 1: @andjsmit The "[0]" goes away if this is a single-valued field, as per f642b02

app/models/solr_document.rb
```diff
@@ -31,6 +31,7 @@ class SolrDocument
   attribute :num_collections, Solr::String, solr_name('num_collections', :stored_sortable, type: :integer)
   attribute :num_works, Solr::String, solr_name('num_works', :stored_sortable, type: :integer)
   attribute :holding_location, Solr::String, solr_name('holding_location')
+  attribute :rights_note, Solr::String, solr_name('rights_note')
```
aploshay, line 34: @andjsmit if single-valued,this needs a 2nd argument, such as 
Solrizer::Descriptor.new(:text, :stored, :indexed)
or
Solrizer::Descriptor.new(:text_en, :stored, :indexed)

to produce a suffix of _tsi or _tesi instead of _tesim

app/views/records/show_fields/_rights_note.html.erb
aploshay: @andjsmit we probably also want to wrap this in a simple_format() or simple_format(auto_link()) call, is done for "description":
https://github.com/samvera/hyrax/blob/v2.9.6/app/views/hyrax/collections/_collection_description.erb#L2

andjsmit: @aploshay Removed index. I did not ad simple_format as this messed display up by adding unwanted p and br tags and AFAIK this is not meant to be a link.

app/models/solr_document.rb
andjsmit: Ok - Leaving as is.

app/views/records/show_fields/_rights_note.html.erb
aploshay: @andjsmit are p, br tags unwanted if they are reflecting any newlines in the textbox value?  @nhomenda?

andjsmit: @aploshay @nhomenda My concern was seeing an extra <p> and <br> tag being added to the content of the field. Turned out these extra tags came from pasting content into the field not as a result of wrapping the field. I've updated PR with simple_format wrap.

rwingiupui:
+1 Desk Checked - Appears to be working properly.

aploshay (approved):

rwingiupui:
+1 Desk Checked - Appears to be working properly.

PR #478, opened 2022-09-13
[ESSI-1622] add campus collection breadcrumbs to catalog results

aploshay:
The first couple of commits copy and modify hyrax's `NestingIndexAdapter` to add an additional property.
* It would probably be more elegant to monkeypatch the hyrax version, but I ran into repeated problems doing that, and reverted to a strict override
* Note that here's active work on deprecating `NestingIndexAdapter` in future versions of hyrax
  * At some future point, we can slim down the essi version to just the portion we're still using, or otherwise refactor
* Note that works and collections will need to be re-saved to update their solr indices, to provide values for the UI components

3rd commit adds UI components to render the breadcrumbs

The next couple of commits copy and modify a hyrax partial -- there are distinct partials to display collection thumbnail and work thumbnails, and the collection-specific version was missing a style class necessary to prevent the breadcrumbs from appearing uncomfortably closely below the thumbnail, in a context of sparse collection metadata that doesn't naturally push the next div lower.

The final commit effectively excludes campus from the catalog metadata display, by changing the custom renderer for campus to always return an empty string, and changing the field display logic to skip over fields if the value (most likely from a custom renderer) ends up being blank.


aploshay:
(re-pushed since I forgot to include the updated spec in the final commit)

dlpierce (approved):

PR #480, opened 2022-09-23
[ESSI-1692] dynamic manifest metadata

aploshay:
Adds allinson_flex metadata fields in the IIIF manifest.

Adds handling of campus as another specially-rendered field (along with holding_location), and fixes display on the work show page to use the campus label (rather than internal code value).


randalldfloyd (approved):
This desk checks for me locally - I get campus/holding location in both the manifest and the UV.

PR #481, opened 2022-09-26
ESSI-1698 Adds okcomputer check to compare image commit levels

randalldfloyd:
Since Sidekiq is likely to run in a separate container, possibly on a separate server, it is important to have some way of making sure that it is running the same commit level as the web container.  This adds a check to OK Computer that compares commit levels from image data extracted during the deployment process.  

It is a bit opinionated in that it assumes the existence of certain files created by running `podman image inspect` on the web and job  images.  Those files are visible from either container due to their location in the shared secrets directory, which allows this status check to open them and compare the environment variable SOURCE_COMMIT.

The status check is only registered if it detects a certain flag file in the secrets directory.  This allows easily configuring this check to occur per instance instead of having to coordinate whether or not keys exist in the main secrets configuration file.

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -42,6 +42,8 @@ Rails.application.configure do
     checks.register "fcrepo", EssiDevOps::FcrepoCheck.new(fcrepo_url, 10,
                                                         auth_options)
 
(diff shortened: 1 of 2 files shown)


aploshay (approved):
+1 by inspection

PR #484, opened 2022-10-07
ESSI-1542 Update Ruby, Node, FITS and some gems

dlpierce:
Updates the versions of Ruby, Node, FITS and some gems.

I've been using the updated Ruby and Node for a while now and haven't noticed any breakage. The issue that prevented us from using 2.7.5 seems to not be a problem in 2.7.6 despite not being able to update to the latest oai gem that had a fix.


aploshay (approved):
Looks good and the new build spins up on my local

rwingiupui:
+1 Desk Checked - No issues observed.

PR #488, opened 2022-11-07
ESSI-1724 Overriding ImageMagick policy.xml to fix large file failures

randalldfloyd:
Large files were failing during derivative creation due to MiniMagick hitting memory limits. The default ImageMagick policies for resources in the base image are too restrictive, so this PR overwrites the policy definitions (in `config/ImageMagick-6/policy.xml`  with a stashed version that increases resource values for memory and dimensions. 

You don't need to ingest a large file to see if this works in the resulting image build.  Get a shell in the container and use  `convert` to list the resource limits.  It should produce the following, which is known to work both locally and on the server:
```
$ convert -list resource
Resource limits:
  Width: 32KP
  Height: 32KP
  List length: unlimited
  Area: 128MP
  Memory: 5GiB
  Map: 1GiB
  Disk: 5GiB
  File: 12288
  Thread: 4
  Throttle: 0
  Time: unlimited
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -67,6 +67,10 @@ RUN gem update bundler && \
 
 COPY --chown=essi:essi . .
 
+# The defaults for ImageMagick are too constrained, override so that MiniMagick won't fail
+RUN mkdir -p /etc/ImageMagick-6/
+COPY ./config/ImageMagick-6/policy.xml /etc/ImageMagick-6/
+
 ENV RAILS_LOG_TO_STDOUT true
 ENV RAILS_ENV production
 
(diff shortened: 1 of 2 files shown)


dlpierce:
I see the height and width limits have been increased, but is the area limit of 128MP an appropriate size for our usual images? Going over the limit just means disk cache is used instead of memory, but presumably there is a significant speed penalty.

randalldfloyd:
Area was adjusted to at least match what width x height would allow for..

dlpierce (approved):

PR #491, opened 2022-11-15
[ESSI-1693] ensure UV display on screens below 640px width

aploshay:


wang58 (commented):
The change is straightforward and I am good with it. But I'd like to know 390px is from hyrax or the number you picked.

aploshay:
@wang58 the 390px value is specified in newer versions of hyrax:
https://github.com/samvera/hyrax/blob/3.x-stable/app/assets/stylesheets/hyrax/_viewer.scss#L3

wang58 (approved):
great!

PR #494, opened 2022-11-18
[ESSI-1672] Remove Slideshow

andjsmit:
Simple config change to remove slideshow from collection index and collection home pages.


rwingiupui (approved):
+1 Desk Checked - Slideshow view no longer available.  No issues observed.

PR #496, opened 2022-11-30
[ESSI-588] refactor manifest viewing hint

aploshay:
The ESSI-588 monkeypatch to add page-level viewingHint values in the manifest was:
1.  ineffective, due to inclusion via `include` instead of `prepend`
2. inefficient, due to running a solr query to look up the `FileSet` `viewing_hint` property, when that's available from the already-built `FileSetPresenter` that `CanvasBuilder` has access to when calling `image_builder`

This PR should resolve both issues.

Note that the current UV does not seem to actually file-specific viewing_hint properties; presumably a future version may.


wang58 (approved):
The change looks good to me. I like your approach on removing the solr query and using the already-built object and method. Also using prepend instead of include is good to me.

I tested viewing hint feature in File manager, it worked great. Nice work!

PR #501, opened 2023-01-06
[ESSI-1765] have CharacterizeJob optionally unlink characterization file, parent dir

aploshay:
File ingest, in the case of both `store_files?` and `transform_to_jp2?` being true, was leaving a .jp2 file around indefinitely post-ingest, in the pattern of:
`/tmp/{tmpdir}/{original file}.jp2`
e.g.
`/tmp/d20221230-1-zz7hny/VAB8235-U-00446-00002.jp2`

This commit adds an argument to trigger cleanup at the end of `CharacterizeJob`, when the .jp2 is no longer needed.  (The `CreateDerivativesJob` called next uses the original .tiff file.)

### DIFF
--- a/app/jobs/characterize_job.rb
+++ b/app/jobs/characterize_job.rb
@@ -6,6 +6,7 @@ class CharacterizeJob < ApplicationJob
   # @param [FileSet] file_set
   # @param [String] file_id identifier for a Hydra::PCDM::File
   # @param [String, NilClass] filepath the cached file within the Hyrax.config.working_path
+  # @param [FalseClass, TrueClass, String, NilClass] delete_characterization_path triggers deleting the file on any truthy value, and the parent directory on 'include_parent_dir'
   def perform(file_set, file_id, filepath = nil, derivation_path = nil, delete_characterization_path = false)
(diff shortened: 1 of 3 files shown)


app/jobs/characterize_job.rb
```diff
@@ -6,16 +6,28 @@ class CharacterizeJob < ApplicationJob
   # @param [FileSet] file_set
   # @param [String] file_id identifier for a Hydra::PCDM::File
   # @param [String, NilClass] filepath the cached file within the Hyrax.config.working_path
```
dlpierce, line 8:
```suggestion
  # @param [String, NilClass] filepath the cached file within the Hyrax.config.working_path
  # @param [String, NilClass] delete_characterization_path cleanups parent directory when set to 'include_parent_dir'
```

dlpierce (approved):
Looks good. I suggested documenting the additional job argument.

app/jobs/characterize_job.rb
aploshay: @dlpierce Oh, snap, I skipped that!  Addressed in new commit

PR #503, opened 2023-01-10
Essi-1762 Add minio container

dlpierce:
Add minio container for testing of S3 storage. Remove BE provider for s3proxy.


aploshay (approved):
+1, everything spins up locally via docker-compose

PR #513, opened 2023-02-21
Essi-1780 errbit config

dlpierce:
Adds config options for using errbit as the target for airbrake error reports


aploshay (approved):
+1 by inspection

PR #514, opened 2023-02-22
[ESSI-1786] add new supplemental holding location

aploshay:


rwingiupui (approved):
+1 - Desk Checked - New holding location now available in the "Holding Location" drop down list.

PR #515, opened 2023-02-23
[ESSI-1786] holding location rendering: handle missing contact values

aploshay:
Rendering code assumes both email and telephone are provided.  This revision should properly handle either/neither/both values being present.


rwingiupui (approved):
+1 Desk Check - All holding information now visible - looks good!

PR #516, opened 2023-02-27
[ESSI-1750] support s3 connections patched for HCP

aploshay:
While it does appear that a blank header value for `accept-encoding`  should be treated equivalently to an `identity` value, the Hitach HCP S3 server isn't doing that.  So this monkeypatches the aws gem to explicitly provide the accepted header value.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -110,3 +110,6 @@ gem 'stackprof', require: false
 
 # hold back hydra-head due to implicit Blacklight 7 requirement in newer versions
 gem 'hydra-head', '10.6.1'
+
+# s3 support for HCP
+gem 'aws-sdk-s3'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1070,6 +1070,7 @@ PLATFORMS
 DEPENDENCIES
   airbrake
   allinson_flex!
+  aws-sdk-s3
   bagit
   better_errors
   binding_of_caller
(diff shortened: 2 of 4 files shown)


dlpierce:
Is it not possible to set this header when creating the request, instead of changing the global default?

aploshay:
@dlpierce we could monkeypatch this differently, but I don't think we can get out of doing some fairly sweeping version of this, unless/until they fix handling on the Hitachi side (as Randall Embry has requested of Hitachi) in a new release, to handle "" and "identity" values equivalently, as they are supposed to but don't do currently.

Running through the example code that I have attached to the story in JIRA:
#list_objects calls #build_request, which builds a new `Seahorse::Client::RequestContext` instance -- and that _could_ be passed an explicit `Http::Request` object (instead of [defaulting to building a generic, blank instance](https://github.com/aws/aws-sdk-ruby/blob/e24391ae2ea5ef0961a9253510042faf3af898a2/gems/aws-sdk-core/lib/seahorse/client/request_context.rb#L7-L28)), which _could_ be [provided with specific header values](https://docs.aws.amazon.com/sdk-for-ruby/v2/api/Seahorse/Client/Http/Headers.html) -- but then it looks like #build_request would have to be patched to accept (a) new optional argument(s), and we'd have to patch every method that could could call #build_request?

dlpierce (approved):

PR #517, opened 2023-03-01
[ESSI-1787] remove incomplete, invalid holding location API results

aploshay:


rwingiupui (approved):
+1 Desk Check: Changes remove blank holding locations from list

PR #527, opened 2023-04-06
essi-1807 Allow Sidekiq queue to be set from ENV

dlpierce:
This will enable us to run additional sidekiq processes dedicated to specific queues


aploshay (approved):
+1 by inspection

PR #531, opened 2023-04-14
Catalog to uv search

kirkkwang:
ref: https://github.com/scientist-softserv/iiif_print/issues/183

Video demo: https://share.getcloudapp.com/RBurr4bA#

[Remove files that are covered by IIIF Print](https://github.com/IU-Libraries-Joint-Development/essi/commit/9e1539748362b1b08e0c345b791e7f7afc2883ad) 

This commit will remove iiif_helper, _index_header_list_default, and
_universal_viewer. These files are covered by the IIIF Print gem and
would conflit if they were left in the application.  This added a
configuration to the initializer to set the UV config path to account
for `uv_config.json` (with an underscore and not a -).  Lastly, we had
to change the config/uv/uv.html to account for the removal of the
javascript and jquery that was in the _universal_viewer partial to
maintain the same functionality of doing an automatic search from the
catalog to UV.


laritakr (approved):
Looks great!

PR #533, opened 2023-04-19
WIP Adds iiif_print to ESSI

randalldfloyd:


Dockerfile
```diff
@@ -67,12 +73,9 @@ RUN gem update bundler && \
 
 COPY --chown=essi:essi . .
 
-# The defaults for ImageMagick are too constrained, override so that MiniMagick won't fail
```
randalldfloyd, line 70: Does the new ImageMagick version installed for iiif_print have policy defaults above what this was intended to address?

app/controllers/catalog_controller.rb
```diff
@@ -18,11 +18,13 @@ def self.modified_field
   include BlacklightIiifSearch::Controller
 
   configure_blacklight do |config|
+    # IiifPrint index fields
+    config.add_index_field 'all_text_tsimv', highlight: true, helper_method: :render_ocr_snippets
```
randalldfloyd, line 22: What does this do?  Will this add context snippets to catalog search results views, or is this related somehow to UV search term highlighting?

app/controllers/concerns/essi/works_controller_behavior.rb
```diff
@@ -23,16 +23,16 @@ def additional_response_formats(wants)
     # Overrides stock Hyrax method to accept retrieving a cached JSON manifest_builder
     def manifest
       headers['Access-Control-Allow-Origin'] = '*'
-  
-      json = sanitize_manifest(JSON.parse(manifest_builder))
-  
+      # manifest builder is now being handled in the IIIF Print gem
+      json = iiif_manifest_builder.manifest_for(presenter: presenter)
```
randalldfloyd, line 27: If the manifest builder is changed to be handled by iiif_print, will it still cache manifests like or similar to the custom `manifest_builder` method?

Dockerfile
randalldfloyd: The policies seem to be using the available resource strategy to set limits. This is what is reported from within the container for the test instance, which is a lot bigger than the explicit values we had set:
```
convert -list resource

Resource limits:
  Width: 461.169PP
  Height: 461.169PP
  Area: 134.527GP
  List length: unlimited
  Memory: 62.6438GiB
  Map: 125.288GiB
  Disk: unlimited
  File: 12288
  Thread: 4
  Throttle: 0
  Time: unlimited
  ```

File: app/controllers/concerns/essi/works_controller_behavior.rb (hunk 3 above)
Comment (randalldfloyd) [reply app/controllers/concerns/essi/works_controller_behavior.rb:27]: Also, what were we fixing by also running the resulting json through [custom sanitize methods](https://github.com/IU-Libraries-Joint-Development/essi/blob/main/app/controllers/concerns/essi/works_controller_behavior.rb#L58-L72)?  That has also been undone by the direct call to the iiif_print manifest builder.

Comment (randalldfloyd) [reply app/controllers/concerns/essi/works_controller_behavior.rb:27]: Info on how to implement Hyrax manifest caching feature has been captured in https://iu-uits.atlassian.net/browse/ESSI-1847 .  

But we should still think about any effects of losing the custom sanitize methods as noted above.

File: app/jobs/create_ocr_job.rb [hunk 4]
```diff
@@ -8,7 +8,9 @@ def perform(file_set, file_id, filepath = nil)
     return if file_set.video? && !Hyrax.config.enable_ffmpeg
     filename = Hyrax::WorkingDirectory.find_or_retrieve(file_id, file_set.id, filepath)
 
-    file_set.send(:file_set_derivatives_service).send(:create_ocr_derivatives, filename)
+    # using #create_derivatives instead of #create_ocr_derivatives to use IIIF Print
+    # @see https://github.com/scientist-softserv/iiif_print/blob/d14246664048c708071c7ff4de2e9a34aa703465/app/services/iiif_print/pluggable_derivative_service.rb#L25
+    file_set.send(:file_set_derivatives_service).send(:create_derivatives, filename)
```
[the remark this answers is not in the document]
randalldfloyd: @andjsmit I'm sure that's probably the case, unless the iiif_print pluggable derivative stack for a particular work type doesn't have any actual steps configured to run.  I wonder how often the recreate job is actually run?

PR #539, opened 2023-05-05
Improve Selenium configuration and improve multiplatform support

dlpierce:
This started out as a selenium update, but became "Make feature specs work on an ARM cpu"

- Update selenium and capybara
- Use seleniarm chromium image for ARM support
- Respect platform when setting up jemalloc
  - _This one had an innocuous warning at rails startup, but then caused Fedora to respond with a 500 because that "harmless" warning was being included as part of the mime type returned by the characterization tools._
- Set chromium resolution from ENV to workaround broken support for doing it from the spec.
- Add spec/fixtures to browse everything for testing in docker.


kirkkwang (approved):
Awesome, thanks for the improvements!

PR #541, opened 2023-05-11
[ESSI-1805] [ESSI-1812] ingest prederived characterization

aploshay:
Follows the same pattern we have for using pre-derived OCR:

1. Look for a pre-derived file, in the configured directory, with a filename matching the original but substituting "-characterization.xml" for the original file suffix
2. If found, use that file; if not, revert to original behavior (generating characterization via the hyrax stack)

Third commit:
1. refactors into a module, code that was common across the two pre-derivation cases (file characterization, OCR)
2. allows for both IUPUI and IUB style of specifying a derivation folder -- a single folder for everything, vs nested work_id/type subfolders

Fourth commit changes the `skip_derivatives` setting from (when true) doing a "hard skip" of OCR (neither generating it, nor reading pre-provided) to a "soft skip" (using pre-derived when provided, but not generating if nothing was pre-derived)

### DIFF
--- a/app/helpers/prederivation_helper.rb
+++ b/app/helpers/prederivation_helper.rb
@@ -1,0 +1,44 @@
+module PrederivationHelper
+  def work_root_name(filename)
+    filename.split('/').last.split('-').first
+  end
+
+  def derivatives_folder_for(filename, type: '')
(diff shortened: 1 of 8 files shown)


dlpierce (approved):
:+1: Looks good by inspection.

PR #543, opened 2023-05-15
Essi-1809 derivatives/imageproc queue; Sidekiq 7

dlpierce:
Update to Sidekiq 7 which required refactoring of our configuration.
CharacterizationJob is sent to the 'derivatives' queue, and CreateDerivativesJob follows suit.
A dedicated sidekiq worker process is setup to allow parallel processing.


aploshay (approved):
+1

PR #544, opened 2023-05-15
Update to ruby 2.7.8

dlpierce:


rwingiupui (approved):
+1 Desk Checked - Checked locally via Docker, no issues observed.

PR #545, opened 2023-05-15
System dep updates

dlpierce:
Update FITS to 1.6.0 and update the CircleCI config to match our docker environment


rwingiupui (approved):
+1 Desk Check - Tested via Docker - No issues observed.

PR #547, opened 2023-05-30
ESSI-1819 Disable FFIdent

dlpierce:
Prevents some JP2000 files being identified as MS Word files


randalldfloyd (approved):
Looks right on inspection; didn't pull down to desk check.

PR #550, opened 2023-06-09
Essi-1815 list source bloat

dlpierce:
This should fix the problem of list_source accumulating old resources when adding a new file to a work (or until recently, just saving the work without adding files).

~~Still need to find a way to identify and cleanup our existing bloated list_source data...~~
Now includes a clean up job.


aploshay (approved):

aploshay (approved):
+1 re-approval of updated PR with cleanup job

PR #554, opened 2023-06-28
attempt to load missing ocr_state property during indexing

aploshay:


aploshay:
deprecating in favor of #555

PR #555, opened 2023-06-29
ESSI-1837 Load all work classes at startup

dlpierce:
Create transient new objects for all registered work types. Attempts to work around failures to define methods for some properties when loading works (e.g. missing ocr_state)


aploshay (approved):

dlpierce:
Had to avoid loading classes when the db tables have not been migrated/created yet. I do wonder if this will cause this fix to not be active when running spec tests?

PR #569, opened 2023-09-25
Improvements to indexing jobs to handle large scale reindexing operations

randalldfloyd:
### Parametrizes the nested indexer depth for control during reindexing
When reindexing all works and collections, we want to control the depth of the nested indexer to just the level of the objects being acted upon. The depth of the nested indexer is set in an initializer, which was hardcoded at 5 but is now configurable in secrets as `essi, :nested_indexer_depth` .  

NOTE that anytime this is set lower than 3 to 5, it is nearly guaranteed that the depth limit will be exceeded and will raise an exception, so it must be handled properly if the goal is to merely skip child indexing.  See the modified batch indexers also in this PR.  Lowering the limit should be done deliberately when the indexing code or caller is aware of the limit.

### Refactors re-indexing jobs to be based on Solr queries
The re-indexing jobs for works, FileSets, and Collections query Solr for documents that have yet to be updated for the purposes of any given re-index operation.  The field name and the value being checked can be passed into the `perform` method, but defaults are taken from configuration (described below.)  Other options can be passed in, such as the row count to control batch sizes and an optional sleep break duration between batches.  The works job can additionally be told whether or not to index member FileSets (via the new job added here) and whether or not to async those jobs. 

As they currently stand, once the jobs are queued they are intended to loop until there are no candidate documents per the given query.  Testing might bear out the need to control the amount of work in a single job run or when it is allowed to run.

### Adds a versioning field to Solr documents to track indexing status
This PR adds to the actual indexer classes to make them responsible for populating the field being checked, versus setting the value in the jobs.  Updating objects from within the jobs will indirectly result in the value being set, and creating/saving objects will also continue to set the value going forward for consistency. 

### Addresses large redundant Solr fields causing performance problems
Development of the reindexing jobs helped uncover a big problem hurting Solr response performance.  The queries to gather candidate documents were taking a very long time to return, which was tracked down to the Solr response writer streaming back the multiple redundant fields containing the full Alto OCR.  This PR removes one of those redundant fields that ESSI code introduced but is no longer needed.  But more importantly, it reduces the value set in the remaining fields to that of only the extracted text and not the full Alto OCR.

### DIFF
--- a/app/jobs/reindex_file_sets_job.rb
+++ b/app/jobs/reindex_file_sets_job.rb
@@ -1,11 +1,45 @@
 # frozen_string_literal: true
 
-# Triggers reindexing on all FileSets
 class ReindexFileSetsJob < ApplicationJob
-  def perform
-    FileSet.find_each do |fs|
-      Rails.logger.info "Reindexing #{fs.class}: #{fs.id}"
-      fs.update_index
+  def perform(file_set_ids = [], opts = {})
+    index_query_field = opts.dig(:index_query_field) || 'iiif_index_strategy_tesim'
+    index_query_value = opts.dig(:index_query_value) || IndexerHelper.iiif_index_strategy
+    row_count = opts.dig(:row_count) || 500
+    sleep_break = opts.dig(:row_count) || 10
+    if file_set_ids.empty?
+      index_query = "-#{index_query_field}:#{index_query_value}"
+      model_query = "has_model_ssim:FileSet"
+      loop do
+        results = ActiveFedora::SolrService.instance.conn.get(
+          ActiveFedora::SolrService.select_path,
+          params: { qt: "search", q: [model_query], fq: [index_query], fl: "id", rows: row_count}
+        )
+
+        break if results["response"]["numFound"] == 0
+
+        results["response"]["docs"].each do |solr_hit|
+          index_file_set(solr_hit["id"])
+        end
+        # Give the backend and the jobs a break...
+        sleep sleep_break
+      end
+    else
+      file_set_ids.each do |id|
+        index_file_set(id)
+      end
     end
   end
+
+  private
+
+    def index_file_set(id)
+      begin
+        fs = FileSet.find(id)
+        FileSet.new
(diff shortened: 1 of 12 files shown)


app/jobs/reindex_works_job.rb
```diff
@@ -1,12 +1,41 @@
...
 class ReindexWorksJob < ApplicationJob
-  def perform
+
+  def perform(opts = {})
+    index_query_field = opts.dig(:index_query_field) || 'iiif_index_strategy_tesim'
+    index_query_value = opts.dig(:index_query_value) || IndexerHelper.iiif_index_strategy
+    index_file_sets = opts.dig(:index_file_sets) || true
+    async_fs_jobs = opts.dig(:async_fs_jobs) || false
+    row_count = opts.dig(:row_count) || 100
+    sleep_break = opts.dig(:row_count) || 10
```
dlpierce, line 11:
```suggestion
    sleep_break = opts.dig(:sleep_break) || 10
```

dlpierce (changes requested):

app/jobs/reindex_file_sets_job.rb
```diff
@@ -1,11 +1,45 @@
...
 class ReindexFileSetsJob < ApplicationJob
-  def perform
-    FileSet.find_each do |fs|
-      Rails.logger.info "Reindexing #{fs.class}: #{fs.id}"
-      fs.update_index
+  def perform(file_set_ids = [], opts = {})
+    index_query_field = opts.dig(:index_query_field) || 'iiif_index_strategy_tesim'
+    index_query_value = opts.dig(:index_query_value) || IndexerHelper.iiif_index_strategy
+    row_count = opts.dig(:row_count) || 500
+    sleep_break = opts.dig(:row_count) || 10
```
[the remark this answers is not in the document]
randalldfloyd: New commits force pushed to resolve above.

app/jobs/reindex_works_job.rb
randalldfloyd: New commits force pushed to resolve above.

dlpierce (approved):

PR #570, opened 2023-10-03
[ESSI-1879] nested filesets in iiif manifests

aploshay:
It turns out that [IIIFManifest::ManifestBuilder::DeepCanvasBuilderFactory#file_set_presenters](https://github.com/samvera/iiif_manifest/blob/v0.5.0/lib/iiif_manifest/manifest_builder/deep_canvas_builder_factory.rb#L7) calls a DeepFileSetEnumerator that [explicitly lists all direct member FileSets, _then_ loops through member works](https://github.com/samvera/iiif_manifest/blob/v0.5.0/lib/iiif_manifest/manifest_builder/deep_canvas_builder_factory.rb#L18-L25).  So, that explains why we're seeing exactly that behavior.

These commits provide a proof-of-concept change to produce the expected and desired behavior of FileSet descendants getting listed in the manifest sequentially, regardless of their depth in the tree. 

Unrelated failing test is resolved via #587 
Needs view partial fix provided by #588 to properly render a mix of nested work/fileset objects

### DIFF
--- a/lib/extensions/iiif_manifest/manifest_builder/deep_file_set_enumerator/nested_each.rb
+++ b/lib/extensions/iiif_manifest/manifest_builder/deep_file_set_enumerator/nested_each.rb
@@ -1,24 +1,32 @@
-# unmodified from iiif_manifest
+# modified from iiif_manifest
 module Extensions
   module IIIFManifest
     module ManifestBuilder
       module DeepFileSetEnumerator
         module NestedEach
-          # lists all direct file_set members, then recurs through work members
+          # modified from iiif_manifest
+          # lists all direct file_set members and member works' file_sets, interspersed
           def each(&block)
-            file_set_presenters.each do |file_set_presenter|
-              yield file_set_presenter
-            end
-            work_presenters.each do |work_presenter|
-              self.class.new(work_presenter).each(&block)
+            member_presenters.each do |member_presenter|
+              if member_presenter.id.in? file_set_presenters.map(&:id)
+                yield file_set_presenters.find { |fp| fp.id == member_presenter.id }
+              elsif member_presenter.id.in? work_presenters.map(&:id)
+                self.class.new(work_presenters.find { |wp| wp.id == member_presenter.id } ).each(&block)
+              end
             end
           end
 
           private
+            # modified from iiif_manifest with memoization
             def file_set_presenters
               @file_set_presenters ||= work.try(:file_set_presenters) || []
             end
 
+            def member_presenters
+              @member_presenters ||= work.try(:member_presenters) || []
+            end
+
+            # modified from iiif_manifest with memoization
             def work_presenters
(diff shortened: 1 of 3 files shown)


lib/extensions/iiif_manifest/manifest_builder/deep_file_set_enumerator/nested_each.rb
```diff
@@ -1,18 +1,25 @@
...
-            file_set_presenters.each do |file_set_presenter|
-              yield file_set_presenter
-            end
-            work_presenters.each do |work_presenter|
-              self.class.new(work_presenter).each(&block)
+            member_presenters.each do |member_presenter|
+              if member_presenter.id.in? file_set_presenters.map(&:id)
+                yield file_set_presenters.find { |fp| fp.id == member_presenter.id }
+              elsif member_presenter.id.in? work_presenters.map(&:id)
+                self.class.new(work_presenters.find { |wp| wp.id == member_presenter.id } ).each(&block)
```
dlpierce, line 13: Are the multiple calls to `file_set_presenters` or `work_presenters` going to cause repeated builds of the presenter objects or their solr docs?

aploshay: It shouldn't -- the method ultimately called should only build them the first time, via the nil guard:
https://github.com/samvera/iiif_manifest/blob/v0.5.0/lib/iiif_manifest/manifest_builder/deep_canvas_builder_factory.rb#L29-L31
https://github.com/samvera/hyrax/blob/v2.9.6/app/presenters/hyrax/work_show_presenter.rb#L176
https://github.com/samvera/hyrax/blob/v2.9.6/app/presenters/hyrax/member_presenter_factory.rb#L29-L37

I did have another version of this loop that was more concisely written and should have been logically equivalent, but it seemed to misbehave spookily, so I reverted to this less elegant version.

dlpierce: It's meant to work that way, but I think [this line](https://github.com/samvera/hyrax/blob/v2.9.6/app/presenters/hyrax/work_show_presenter.rb#L280-L282) that creates a new factory object makes the memoization in the factory useless.

aploshay: Oh, snap!  What about patching that line to the following?
```ruby
      def member_presenter_factory
        @member_presenter_factory ||= MemberPresenterFactory.new(solr_document, current_ability, request)
      end
 ```

dlpierce: Seems reasonable

randalldfloyd (approved):

PR #576, opened 2023-10-27
More improvements to reindexing jobs

randalldfloyd:
Based on trial and error, this PR adds more ability and refinements to the reindexing jobs.  It adds more ability to track status and failures in versioned log files, and allows constraining single invocations to a particular work type (useful for parallel processing because it avoids two threads trying to work on the same set of IDs obtained through the Solr queries.)

Also, it was found through testing that FileSets were taking a long time to index due to duplicate LDP load operations in the FileSetIndexer.  Those were being caused by repetitious calls to the FileSet's parent when storing membership details in Solr fields.  This refactors to only access the parent once, which takes the average time from 9 seconds to 3.

### DIFF
--- a/app/indexers/essi/file_set_indexer.rb
+++ b/app/indexers/essi/file_set_indexer.rb
@@ -6,8 +6,11 @@ module ESSI
 
     def generate_solr_document
       super.tap do |solr_doc|
-        solr_doc['is_page_of_ssi'] = object.parent.id if object.parent
-        solr_doc['parent_path_tesi'] = Rails.application.routes.url_helpers.polymorphic_path(object.parent) if object.parent
+        parent = object.parent
+        unless parent.nil?
+          solr_doc['is_page_of_ssi'] = parent.id
+          solr_doc['parent_path_tesi'] = Rails.application.routes.url_helpers.polymorphic_path(parent)
+        end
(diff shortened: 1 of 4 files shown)


app/indexers/essi/file_set_indexer.rb
```diff
@@ -6,8 +6,11 @@ class FileSetIndexer < Hyrax::FileSetIndexer
 
     def generate_solr_document
       super.tap do |solr_doc|
-        solr_doc['is_page_of_ssi'] = object.parent.id if object.parent
```
randalldfloyd, line 9: For reviewers:  This series of calls to the parent resulted in four LDP load operations: once for each parent method invoked but _also_ for each safety check in the `if` clause.

app/indexers/essi/file_set_indexer.rb
```diff
@@ -6,8 +6,11 @@ class FileSetIndexer < Hyrax::FileSetIndexer
 
     def generate_solr_document
       super.tap do |solr_doc|
```
randalldfloyd, line 8: For reviewers: The following calls to the parent resulted in four LDP load operations: once for each parent invocation, but _also_ for each safety check in the if clause.

Storing the parent once in a variable only loads it once but still gets access to the info for the properties.

dlpierce (approved):

PR #580, opened 2023-10-31
[ESSI-1901] Removes roles from Homepage Featured Works/Recently Uploaded list items

andjsmit:


dlpierce (changes requested):
Removing the roles seems like the wrong move if we're improving accessibility. I think the real problem here is `role="tab"` is specified on the \<a> tags which are not the direct child of the element with `role="tablist"`. Can we try moving `role="tab"` to the \<li> tags?

andjsmit:
Got it. Roles returned in new locations.

dlpierce (approved):

PR #581, opened 2023-10-31
[ESSI-1903] Fileset download action non-unique "id" moved to "class"

andjsmit:


dlpierce (approved):
The git history of the addition of `id="file_download"` says it was thought to be needed for google analytics tracking of downloads. I don't think we care if that breaks, at least at this time.

PR #582, opened 2023-10-31
Essi-1902 Essi-1860 Remediation of empty events table

dlpierce:
Fixes essi-1902 by rendering a static table row when the events list is empty.


wang58 (approved):
Looks good to me.

PR #583, opened 2023-11-02
Refactor default pdf_state value to handle absent/single/multi-values

aploshay:


randalldfloyd (approved):
+1 Catching the argument error seems like an elegant solution; otherwise you would have to figure out what the declared type should be from any instance of an ArchivalMaterial and its last saved profile version.

PR #584, opened 2023-11-16
Changes striped table link color to IU Dark Midnight 

andjsmit:
Fixes color contrast issue for striped tables. Should now pass AA and AAA standards, plus part of IU color scheme.


rwingiupui (approved):
+1 Desk Check: Checked link color using Chrome developer tool - color contrast using correct contrast ratio

PR #589, opened 2023-12-01
replace #contains_exactly with #contains check to kludge around spec Collection persistence

aploshay:
Another attempt to avoid that flaky test failure


dlpierce (approved):
This should prevent the flakiness. I imagine adding :clean to the spec metadata also would.

PR #590, opened 2023-12-13
Essi-1925 Do the Robot 🤖

dlpierce:
Adds the ability to change the content of robots.txt without redeployment or restart.

Editor access is provided to admins in the settings section of the dashboard.

### DIFF
--- a/app/controllers/robots_controller.rb
+++ b/app/controllers/robots_controller.rb
@@ -1,0 +1,43 @@
+class RobotsController < ApplicationController
+  before_action :authenticate_user!, except: :show
+  before_action :find_robots_txt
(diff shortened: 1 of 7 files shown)


randalldfloyd (approved):
This desk checks for me locally.  After editing the content block, I get the identical rules added and the headers are correct:
```
$ curl -k https://localhost:3000/robots.txt

user-agent: GoogleOther
disallow: /search

user-agent: SemrushBot
disallow: /search

$ curl -Ik https://localhost:3000/robots.txt
HTTP/1.1 200 OK
...
Content-Type: text/plain; charset=utf-8
```

PR #594, opened 2023-12-15
resolve bundler/byebug conflict

aploshay:
Dockerhub build is failing with byebug complaint on `bundle install` running bundler 2.1.4
`gem install bundler` is trying to update to bundler 2.5, which requires a newer version of ruby
This forces an update specifically to bundler 2.4.12, which will successfully run `bundle install`


Dockerfile
```diff
@@ -47,7 +47,9 @@ COPY --chown=essi:essi Gemfile Gemfile.lock ./
 # DEV ONLY - REMOVE LATER
 # COPY --chown=essi:essi vendor/engines/bulkrax /app/vendor/engines/bulkrax
 # COPY --chown=essi:essi vendor/engines/allinson_flex /app/vendor/engines/allinson_flex
-RUN gem update bundler
+# hold back from bundler 2.5 onwards until ruby is updated
+# RUN gem update bundler
+RUN gem install bundler -v 2.4.12
```
dlpierce, line 52:
```suggestion
RUN gem install bundler -v 2.4.22
```

Dockerfile
```diff
@@ -71,7 +73,9 @@ RUN bundle config --global frozen 1
 COPY --chown=essi:essi Gemfile Gemfile.lock ./
 # DEV ONLY - REMOVE LATER
 # COPY vendor/engines/allinson_flex vendor/engines/allinson_flex
-RUN gem update bundler && \
+# hold back from bundler 2.5 onwards until ruby is updated
+# RUN gem update bundler && \
+RUN gem install bundler -v 2.4.12 && \
```
dlpierce, line 78:
```suggestion
RUN gem install bundler -v 2.4.22 && \
```

dlpierce (changes requested):
Shouldn't we use the latest 2.4 release?

aploshay:
> Shouldn't we use the latest 2.4 release?

We should!

dlpierce (approved):

aploshay:
@dlpierce looks like we can use Gemfile syntax to specify `"~> 2.4.22" instead of hardlocking to 2.4.22, sound good as a modification?

dlpierce:
Sounds good, and based on the history it seems likely there will be patch releases.

PR #597, opened 2024-01-12
[ESSI-875] bulkrax 5 upgrade

aploshay:
Upgrades from bulkrax 1.0.2 to 5.5.1.

Work and FileSet ingest is working in my local environment, with some outstanding issues.

Known Issues:
* ** ~FileSet <-> Work association**:  The various import jobs aren't being sequenced to run in the necessary order (at least, in my local environment), so I'm finding I have to run the importer once to get the works created (with all the FileSets failing), then re-run it to create all the FileSets.  See~ ~[bulkrax issue #955]~(https://github.com/samvera/bulkrax/issues/955)~
* ** ~Collection creation/association**: I haven't had success ingesting to create a new Collection, or to associate to an existing one -- but I only tried this lightly, and it might be issues of configuration and/or CSV content.~
* **METS XML import**:  dropped, as it is based on XML import, which isn't supported in newer version of bulkrax
* **Ingested nested Works within Works**: works, but doesn't display in UI, but that's really a separate issue
* **Exporting**: metadata export works, ~but with access concerns~, access restricted to admins and managers, and file export prevented (isn't working, and we don't want it done, anyway)

Temporarily skipped tests:
* Feature spec for CSV ingest
* METS XML Entry model spec (parser is still tested)

Regarding `configuration change` label:
* requires running database migrations
* may also require (or benefit from) TBD config changes to bulkrax

Notes on running locally:
* make sure to run a db:migrate to pick up the bulkrax changes -- the rails server will run without complaint when the migrations are coming from a gem, unlike when they are directly in the repo and it hard stops and tells you to migrate
* you may run into some issues spinning up the branch; ask me if I had the same problem and what I did to work around it


dlpierce (approved):
Looks good by visual inspection.
The downgrade of sidekiq is probably a good thing. While nothing seemed broken, it does warn that sidekiq 7 is not supported on this old stack.

dlpierce (approved):

PR #602, opened 2024-02-14
[ESSI-1939] fix PDF generation for image widths below 1024

aploshay:
PDF generation currently hardcodes an image width of 1024.  Our IIIF image server returns a 403 if you try to render an image at great than 100% width, which breaks the PDF generation that page if the image is too narrow, and broken generation of any page breaks PDF generation for the entire work.

This changes the IIIF request to send a image width value of:
* if the image is at least 1024 wide: 1024, as before
* if the image is narrower: the actual image width

Further details:
* added more error catching and logging
* if image width is _missing_ from indexed characterization results, it attempts to (re-)call characterization -- first directly, just to get the width value and use it, and subsequently spawning the job


randalldfloyd (approved):
Seems right on inspection; didn't pull/test locally, probably easier with knows use cases once deployed to devel.

PR #605, opened 2024-03-08
[ESSI-1947] refactor banner lookup fedora query to solr query, add memoization

aploshay:
Two performance issues addressed here, in the FIleSetPresenter and WorkShowPresenter logic for collection lookup:
* both currently  make Fedora lookups to build a collection presenter, where a solr lookup suffices and is more performant
* a lack of memoization leads to the WorkShowPresenter collection lookup getting called twice, where once should suffice.

### DIFF
--- a/lib/extensions/hyrax/work_show_presenter/collection_banner.rb
+++ b/lib/extensions/hyrax/work_show_presenter/collection_banner.rb
@@ -3,10 +3,11 @@ module Extensions
     module WorkShowPresenter
       module CollectionBanner
         def collection
-          # return work collection if any, else nil
+          return @collection if @collection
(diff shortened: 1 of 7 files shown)


lib/extensions/hyrax/work_show_presenter/collection_banner.rb
dlpierce, line 1: This is a good improvement! I wonder if using `WorkShowPresenter#member_of_collection_presenters` would be better for finding the collections?

aploshay: @dlpierce see 2nd commit for pertinent changes

dlpierce (approved):
Looks good! :+1:

PR #607, opened 2024-03-18
[ESSI-1949] add members via a single save

aploshay:
Saves once after adding all new ordered_members, rather than once per item.


dlpierce (approved):
This is just moving the save call, looks good to me.

PR #612, opened 2024-04-08
[ESSI-1928] Hidden content focusable

andjsmit:
This PR will make a noticeable change to the home page.

1. Moves "View All Collections" Button up to just below Campus Collection options.
2. Removes Featured Works and Recently Uploaded buttons
3. Displays Featured Works - if any - as left column on large screens
4. Displays Recently Uploaded as right column on large screens if Featured Works
5. Both sections expand to full width on medium screens and smaller 
6. Adds alt text to works thumbnails in both sections

Resolves SiteImprove Issues:

- Role with implied hidden content has keyboard focus
- Image missing a text alternative

### DIFF
--- a/app/assets/stylesheets/digital_collections/_layout.scss
+++ b/app/assets/stylesheets/digital_collections/_layout.scss
@@ -232,6 +232,11 @@ a:hover, a:focus {
   margin: auto;
 }
 
+.home_all_collections {
+  display: table;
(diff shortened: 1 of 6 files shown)


rwingiupui (approved):
+1 Desk Check and SiteImprove browser plugin: Looks good!  All PR changes verified, SiteImprove detected no A or AA issues.

PR #615, opened 2024-04-16
Adding a global controller action to log request details

randalldfloyd:
Adds an `around_action` to the global application controller to log critical request details, like origin IP and agent.  These are essential for mitigating clients exhibiting excessive access behaviors.

The resulting log line will be tagged with a unique label that can be easily searched for in log data:
```
ACCESS: 127.0.0.1, GET https://localhost:3000/, Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/123.0.0.0 Safari/537.36
```


dlpierce (approved):
Code looks good.

If the real client IP is something we want to log here, would this need to include [remote_ip](https://api.rubyonrails.org/v5.2.8.1/classes/ActionDispatch/Request.html#method-i-remote_ip)?

randalldfloyd:
@dlpierce I think I will try to use `remote_ip` instead.  Some say it ends up being the same in practice anyway, but per the code, it actually checks the headers that might contain the first originator ( the `*-FORWARDED-FOR` headers.)  Not all proxies set them, but that may have been what EP meant when they said they should be available.

PR #616, opened 2024-05-13
[ESSI-1927] Accessibility fix. Sets UV maximum-scale to 3.

andjsmit:
According to [MDN Webdocs](https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag), viewport maximum-scale should be set to 3 or higher for accessibility reasons.


rwingiupui (approved):
+1 Desk Check & Code Review:  No issue observed.  Confirmed code change to [config/uv/uv.html] and also verified via Chrome Inspect Tool.

PR #617, opened 2024-06-11
ESSI-1966 Add analytics JS and config

dlpierce:
Override hyrax's _ga.html.erb with what is needed by GA4, and set the id from essi config.


rwingiupui (approved):
+1 Desk Check & View Source: I added a GA id to essi_config.docker.yml file this enabled GA JS code observed via View Source.  No issues observed.

PR #622, opened 2024-08-01
[ESSI 1978][ESSI-1982] Replace IUPUI and IUPUC with IU Indianapolis and IU Columbus

andjsmit:


rwingiupui (approved):
+1 Desk Check: Footer, Image ALT tags and references to IUPUI and IUPUC have been updated.  No issues observed - Looks good!

PR #624, opened 2024-08-21
ESSI-1995 Update IGWS email

dlpierce:


aploshay (approved):
+1 by inspection.  We've got other @indiana.edu email addresses in that file, but those accounts don't appear to be migrated to @iu.edu yet.

PR #633, opened 2024-10-25
[ESSI-2025] campus facet display

aploshay:
For facet display of campus, uses label value (e.g. "IU Indianapolis") for link text, but retains use of code value (e.g. "IUI") for link destination.


rwingiupui (approved):
+1 Desk Check: Facet display of campus name now uses the label value as its link text, no longer using campus code value.  Works as designed.  No issues found.

PR #635, opened 2024-11-12
[ESSI-1918] only show file depositor, activity log to parent work editors

aploshay:


rwingiupui (approved):
+1 Desk Check: Before logging into ESSI (public user), I wasn't able to see the depositor or user activity log for an image.  After logging in, I could see this information.  Works as designed.

PR #638, opened 2024-11-19
ESSI-2029 Configure LDAP to use secure connection

dlpierce:
Use encrypted connection for LDAP connection.

Updates ldap_groups_lookup to a new version that allows direct configuration via initializer instead of needing a yaml file intermediary. This is needed to be able to set tls_options to enable cert verification, and should make configuring the SASL callback easier if that ends up being needed.


aploshay (approved):
Runs great locally!  It looks like there might be a shortly forthcoming 0.10.1 or 0.11.0 release for ldap_groups_lookup?  Should this incorporate that?

PR #644, opened 2025-01-07
[ESSI-2045] Removes depositor info from homepage featured and recent items.

andjsmit:


rwingiupui (approved):
+1 Desk Check: Depositor information has been removed.

PR #646, opened 2025-01-13
[ESSI-2056] fix for downloading extracted_text in external_storage case

aploshay:


dlpierce (approved):
This looks good. I don't think there are other download types to account for?

aploshay:
> This looks good. I don't think there are other download types to account for?

Yeah, I don't think so -- the only other file use we have currently is preservation_master, and that's not exposed in the UI.

PR #648, opened 2025-01-23
[ESSI-2062] fix iiif urls unaware of external_storage

aploshay:
Refactors lookup of IIIF image server URLs (in a manner aware of external_storage settings), and applies that logic where not yet present.

New/modified infrastructure:
* [app/services/iiif_file_set_path_service.rb](https://github.com/IU-Libraries-Joint-Development/essi/pull/648/files#diff-793f5625c32e4b74214afdead845769c64bae3a4bed2a7873646b05b879066f2) (new, replaces direct calls to `Hyrax.config.iiif_image_url_builder`, `Hyrax.config.iiif_info_url_builder`)
* [app/models/file_set.rb](https://github.com/IU-Libraries-Joint-Development/essi/pull/648/files#diff-bac0cfdccf9a69b2827b98b88a6416464fb0313eb425e7be2cfc0a3e04d2b53f) (modified)
* Hyrax::FileSetPresenter [modified via an extension](https://github.com/IU-Libraries-Joint-Development/essi/pull/648/files#diff-b0e14a4db0e2bcfe494d6a4f819e491460711982c4f7363ca8c451b4c0c6e3e0)

Updated cases, adding handling of external storage:
* [app/models/collection_branding_info.rb](https://github.com/IU-Libraries-Joint-Development/essi/pull/648/files#diff-96a6b9dd7c873584c633ad2a985b641a689388b359c744d541333b4ca1af8250) (part of initial bug report generating ESSI-2062)
* [app/services/essi/generate_pdf_service.rb](https://github.com/IU-Libraries-Joint-Development/essi/pull/648/files#diff-92f2bb4d7d1685fafe730ec956c212c22243abce6791e8a9327ee91063c2f5ea) (not tested, but expected to be broken for external storage files)
* [app/services/iiif_collection_thumbnail_path_service.rb](https://github.com/IU-Libraries-Joint-Development/essi/pull/648/files#diff-799b4cbff7f25d6bc34557feb78c0221f06067457357f60d139e1d5612568beb)  (part of initial bug report generating ESSI-2062)

Refactoring cases:
* [app/controllers/purl_controller.rb](https://github.com/IU-Libraries-Joint-Development/essi/pull/648/files#diff-59fc0b2dcdf4a6433bbc3b634ac8ac877a7c7ac4c1071eaf7791593900df58f3)
* [app/helpers/catalog_helper.rb](https://github.com/IU-Libraries-Joint-Development/essi/pull/648/files#diff-d2fc5fb79c7ff4f74a7770150821a67b0ad7e416fc6c82815556268cf417d5dd)
* [app/presenters/hyrax/displays_image.rb](https://github.com/IU-Libraries-Joint-Development/essi/pull/648/files#diff-273f6e2f6358de761e5f1c9918a018e21685c95316b591a7d1f9bf87f2f9ef66)
* [app/services/iiif_thumbnail_path_service.rb](https://github.com/IU-Libraries-Joint-Development/essi/pull/648/files#diff-07cd0c30b1e641a525d123b8230e92eafc248586203173b7f8a9d70f2238344d)

### DIFF
--- a/app/services/iiif_file_set_path_service.rb
+++ b/app/services/iiif_file_set_path_service.rb
@@ -1,0 +1,88 @@
+class IIIFFileSetPathService
+  attr_reader :file_set, :versioned_lookup
+
+  # @param [ActiveFedora::SolrHit, FileSet, Hash, SolrDocument, Hyrax::FileSetPresenter] file_set
+  # @param [TrueClass, FalseClass, NilClass] versioned_lookup: whether to use versioned file lookup if original_file_id fails
+  def initialize(file_set, versioned_lookup: false)
+    @versioned_lookup = versioned_lookup
+    file_set = SolrDocument.new(file_set) if file_set.class.in? [ActiveFedora::SolrHit, Hash]
+    if [:id, :content_location, :original_file_id].map { |m| file_set.respond_to?(m) }.all?
+      @file_set = file_set
+    else
+      raise StandardError, 'Provided file_set does not meet interface requirements' 
+    end
+  end
+
+  def lookup_id
+    @lookup_id ||= (@versioned_lookup ? versioned_lookup_id : basic_lookup_id)
+  end
+
+  # @return [String] a URL that resolves to an image provided by a IIIF image server
+  def iiif_image_url(base_url: nil, size: nil)
+    return unless lookup_id
+    Hyrax.config.iiif_image_url_builder.call(lookup_id, base_url, size || Hyrax.config.iiif_image_size_default)
+  end
+
+  # @return [String] a URL that resolves to an info.json file provided by a IIIF image server
+  def iiif_info_url(base_url)
+    return unless lookup_id
+    Hyrax.config.iiif_info_url_builder.call(lookup_id, base_url)
+  end
+
+  private
+    # imported logic from IIIFThumbnailPathService, etc
+    def basic_lookup_id
+      @basic_lookup_id ||= (file_set.content_location || file_set.original_file_id)
+    end
+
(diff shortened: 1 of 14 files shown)


app/services/iiif_file_set_path_service.rb
dlpierce, line 1: Pulling these IIIF behaviors into this service presents an opportunity to make some spec tests.

aploshay:
@dlpierce for the refactoring pass, I didn't want to _change_ any logic, but there's an opportunity to simplify/unify logic if it was unintentionally variable, originally:
* if we get consistent (either way) about just checking that `content_location` _has_ text (at all) vs starting with "s3", we can simplify #versioned_lookup_id to call #basic_lookup_id
* if we get consistent about checking for file versioning if #basic_lookup_id fails -- originally done only in the `DisplaysImage` module -- we can do away with the #basic_lookup_id vs #versioned_lookup_id split, entirely

aploshay:
@dlpierce so that for failing test, it's complaining that it's not expecting version information in the indexed path:
..**_.%2Ffcr:versions%2Fversion1_**/full/250,/0/default.jpg
vs
.../full/250,/0/default.jpg (expected)

What are your thoughts how to approach a stable fix?
* we could amend the test, so the matching pattern passes with or without version information
* we could amend the test, to _expect_ version information (and fail without)
* we could amend FileSet#original_file_id so it doesn't do any version lookup -- the test should pass as-written, and that might be more in-line with expectations?

dlpierce (approved):
Looking good! Thanks for cleaning this up!

PR #650, opened 2025-02-04
Update webdrivers gem

dlpierce:
Downdate sidekiq (matches same change in bulkrax 5 branch)
Cleanup github git_source
Use posix_spawn fixed for newer system libs.


rwingiupui (approved):
+1 Desk Check: No issues observed.

dlpierce:
I let the marc gem update. Is that okay @aploshay?

aploshay:
@dlpierce marc update seems fine.  I got some other questions/notes:

- What's the effect of having, or not, the `git_source` block?
- I think we might want to change the github source for `willow_sword` from `notch8` to `CottageLabs`, as I'm having trouble bundling locally (directly, non-docker) freshly pulling from the notch8 repo, and the same thing happened the other day on Renee's laptop.  What's your experience?
  - In the bulkrax 5 upgrade, I'm guessing we can just drop the repo, entirely
- Running specs locally, do you get an error on `spec/actors/hyrax/actors/file_actor_spec.rb:51` with respect to looking for a mimetype of "image/png" rather than "message/external-body"?

dlpierce:
I'm going to merge this and we can take care of the other issues in other PRs.

PR #651, opened 2025-02-05
Bundle updates

aploshay:


dlpierce (approved):

aploshay:
@dlpierce I think follow-up work is merited to ensure we can run `bundle update` (without specifying `--conservative` without breaking updates), but this is still some progress and I think its inclusion of `sqlite3` might get us past the current chokepoint on automated builds failing

PR #653, opened 2025-02-05
External body spec

dlpierce:
Adds spec for enabled external storage. We may want to enable external storage for all tests via essi_config.example.yml?


aploshay:
@dlpierce I'm still getting the rspec failure on line 51 running locally in docker; currently unable to run tests locally directly

aploshay (approved):

PR #660, opened 2025-03-12
[ESSI-2075] remove language selector

aploshay:


dlpierce (approved):
:+1: It looks like this also resolves errors for unknown/nonsense locales?

aploshay:
> 👍 It looks like this also resolves errors for unknown/nonsense locales?

Oh, I wasn't specifically intending that -- but, yeah, it does.

Added another commit to consider, adding a 2nd line into `#set_locale` that will fix the browser URL to contain the actually _used_ `locale` on subsequent requests, if it's being coerced.  This feels to me like a reasonable compromise between continuing the display an invalid locale value that's being ignored, vs a hard redirect.

dlpierce (approved):

PR #661, opened 2025-03-25
ESSI-2079 Add rack attack

dlpierce:
Adds rack attack gem and the ability to dynamically control the ip addresses and user agents it acts upon via a yaml block in a text dashboard form.


aploshay (approved):
Looks good to me, the only thing I can think of is that in the instructions, we could link "YAML" to some format documentation somewhere like we're doing for "regular expressions"

PR #671, opened 2025-04-30
[Essi 2093] Add accessibility remediation request and support links 

andjsmit:
Also removes Social Media links from Item and File Set pages.


rwingiupui (approved):
+1 Desk Check & Code Review:  Works as designed, no issues observed.

PR #686, opened 2025-09-03
[ESSI-2130] apply Fedora/S3 file retrieval for OCR generation, characterization, PDF generation, download

aploshay:
Currently, the CharacterizeJob and download action have logic to retrieve files from either Fedora or S3 storage, but corresponding updates are missing from CreateOCRJob and PDF generation.

First commit applies corresponding updates where they're missing; second commit condenses duplicate logic into new methods.


aploshay:
@cjcolvar updated with specs, on current `main`

cjcolvar (approved):
Thanks for adding the tests.  Looks great!

PR #689, opened 2025-09-17
[ESSI-2129] [ESSI-2133] store max height, width characterized

aploshay:
Characterization can _potentially_ produce multiple values for width and height, for pyramidal tiffs.  In the event that it does: we only want the maximal values, so this adds a step in parsing the output of characterization before storing it, to ensure that's what we get.

(Note that I think characterization output from `fits.sh` is _only_ outputting the minimal resolution, but I think we want this change to add robustness when we do start getting the maximal resolution.)

### DIFF
--- a/lib/extensions/hydra/works/characterization_service/precharacterization.rb
+++ b/lib/extensions/hydra/works/characterization_service/precharacterization.rb
@@ -18,7 +18,16 @@ module Extensions
             extracted_md = extract_metadata(content)
           end
           terms = parse_metadata(extracted_md)
+          terms = revise_metadata(terms)
           store_metadata(terms)
+        end
+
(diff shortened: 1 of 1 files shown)


rwingiupui (commented):
Desk Check:
I created a new Works with 5 newspaper pyramidal tiff images also, another Works with 3 pyramidal tiff images (of flowers).  No issues with displaying the above ingested images in the Universal Viewer.

*Possible issue: Each newspaper image had an empty OCR text file.  I regenerated the OCR for each newspaper image, but the OCR file was still empty.
(I think the above issue is the result of my converting an existing tiff image to a pyramidal tiff - maybe the conversion made it less OCR-able?)  Maybe someone else can double-check this with a known/verified pyramidal tiff image?

PR #692, opened 2025-09-26
[ESSI-2129] use higher resolution resolution pdf images

aploshay:
* restores the previous behavior using the _lesser_ of 1024, characterized width (since our IIIF still does seem to raise an error if you ask it to scale above 100%)
* raises the width cap from 1024 to 2048, along with some very light refactoring


cjcolvar (approved):
Looks good to me.

PR #741, opened 2026-07-02
backfill in rack attack controller specs; fix edit page typo

aploshay:
Sending upstream the additions/changes made in projects downstream (datacore, rdc)

### DIFF
--- a/app/views/rack_attacks/edit.html.erb
+++ b/app/views/rack_attacks/edit.html.erb
@@ -4,11 +4,12 @@
 
 <div class="row">
   <div class="col-md-12">
+    <div class="settings_intro">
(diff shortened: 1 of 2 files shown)


cjcolvar:
Looks like the failing build is due to an old version of nodejs.  @dlpierce was talking about how this is due to the image used by circleci is old but that there isn't a newer image available that is still ruby 2.7, the version used by essi.  I'm thinking a path forward might be to switch to using the plain ruby image (instead of `-browsers`) then add the circleci node orb and install a newer version of node that way.  I'll write up a draft PR.

cjcolvar (approved):

PR #742, opened 2026-07-06
Install node ourselves to get a newer version needed for other dependencies

cjcolvar:
Attempt to fix build in main due to newer version of `sass` not working with our old version of nodejs.
FWIW I set this to install nodejs 22.x since it is LTS receiving maintenance until 4/2027 and is the smallest jump to a version not EOL.

See https://github.com/IU-Libraries-Joint-Development/essi/pull/741#issuecomment-4894231450
Possibly related: https://github.com/samvera/hyrax/pull/7475

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,1 +1,2 @@
 defaults
+not op_mob >= 1
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,6 +7,7 @@ version: 2.1
 orbs:
   browser-tools: circleci/browser-tools@1
   ruby: circleci/ruby@2
+  node: circleci/node@7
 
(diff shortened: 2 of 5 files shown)


dlpierce:
Seems like the failing specs are flaky and not tied to the node version.

cjcolvar:
I reverted the last commit to go back to nodejs 22.  Let's see if the failures are any different this time.

cjcolvar:
Failures look sort of similar so trying nodejs 20.20 again.

cjcolvar:
Those failures look the same as the nodejs 22 errors so maybe something is getting cached and not fully regenerated in the build OR the errors are totally independent of the node version.

dlpierce (approved):

PR #743, opened 2026-07-07
[ESSI-2225] add Institution visibility as bulkrax import option

aploshay:


cjcolvar (approved):
Looks good to me!  I'm fine with merging this with the build failing since the failures appear related to the nodejs version which is getting worked on in #742 and I don't think these changes are safe.

PR #745, opened 2026-07-08
Use debian bookworm base image

dlpierce:
Update the base image to debian bookworm.
Install Ruby 2.7.8 with openssl fixes
Update Node to v24
Update Imagemagick to 7.1.2-27
Clean up Dockerfile styling

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,108 @@
 # system dependency image
-FROM ruby:2.7.8-bullseye AS essi-sys-deps
+FROM buildpack-deps:bookworm AS ruby-278-bookworm
+
+# skip installing gem documentation with `gem install`/`gem update`
+RUN set -eux; \
+        mkdir -p /usr/local/etc; \
+        echo 'gem: --no-document' >> /usr/local/etc/gemrc
+
+ENV LANG=C.UTF-8
+
+# Ruby 2.7.8 (note: this version is EOL)
+ENV RUBY_VERSION=2.7.8 \
+    RUBY_DOWNLOAD_URL=https://cache.ruby-lang.org/pub/ruby/2.7/ruby-2.7.8.tar.xz \
+    RUBY_DOWNLOAD_SHA256=f22f662da504d49ce2080e446e4bea7008cee11d5ec4858fc69000d0e5b1d7fb
+
+# OpenSSL 1.1.1 compatibility version for Ruby 2.7.8
+ENV OPENSSL_VERSION=1.1.1w \
+    OPENSSL_DOWNLOAD_URL=https://www.openssl.org/source/openssl-1.1.1w.tar.gz \
+    OPENSSL_DOWNLOAD_SHA256=cf3098950cb4d853ad95c0841f1f9c6d3dc102dccfcacd521d93925208b76ac8
+
+RUN set -eux; \
+        # Compile OpenSSL 1.1.1 from source
+        wget -O openssl.tar.gz "$OPENSSL_DOWNLOAD_URL"; \
+        echo "$OPENSSL_DOWNLOAD_SHA256 *openssl.tar.gz" | sha256sum --check --strict; \
+        mkdir -p /usr/src/openssl; \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,37 +1,127 @@
...
 ARG GROUP_ID=1000
 
 RUN groupadd -g ${GROUP_ID} essi && \
     useradd -m -l -g essi -u ${USER_ID} essi && \
-    curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
-    echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
-    curl -sL https://deb.nodesource.com/setup_16.x | bash - && \
+    curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | gpg --dearmor | tee /etc/apt/keyrings/yarn-archive-keyring.gpg > /dev/null && \
+    echo "deb [signed-by=/etc/apt/keyrings/yarn-archive-keyring.gpg] https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
+    curl -sL https://deb.nodesource.com/setup_24.x | bash - && \
```
cjcolvar, line 100: With this installing nodejs 24 we should probably change the circleci config to use the same version (my PR set it to 20 in the end).  Not a real big deal so I think that it could be in this PR or a separate one.

Dockerfile
```diff
@@ -1,37 +1,127 @@
 # system dependency image
-FROM ruby:2.7.8-bullseye AS essi-sys-deps
+FROM buildpack-deps:bookworm AS ruby-278-bookworm
```
cjcolvar, line 2: There is a lot packed into this stage.  I'm guessing this is basically recreating the ruby 2.7.8-bullseye image for bookworm.  This seems like it could be brittle and finicky to maintain but it makes sense since DC is on old versions of things and on life support until Galacto is done.

Dockerfile
```diff
@@ -1,37 +1,127 @@
...
       yarn libopenjp2-tools libjemalloc2 && \
     apt-get clean all && rm -rf /var/lib/apt/lists/* && \
     ln -s /usr/lib/*-linux-gnu/libjemalloc.so.2 /usr/local/lib/libjemalloc.so.2
 RUN yarn && \
     yarn config set no-progress && \
     yarn config set silent
 RUN mkdir -p /opt/fits && \
-    curl -fSL -o /opt/fits/fits-1.5.5.zip https://github.com/harvard-lts/fits/releases/download/1.5.5/fits-1.5.5.zip && \
-    cd /opt/fits && unzip fits-1.5.5.zip && rm fits-1.5.5.zip && chmod +X fits.sh && \
+    curl -fSL -o /opt/fits/fits-1.6.0.zip https://github.com/harvard-lts/fits/releases/download/1.6.0/fits-1.6.0.zip && \
```
cjcolvar: ImageProc is still on fits 1.5.5.  It seems like it would be advantageous to match versions to avoid odd problems from different behaviors depending on which environment operations were run.  What do you think about reverting this change and doing it later when it can be coordinated with ImageProc?

cjcolvar (commented):
Nice work!  I have a question and a couple comments.

Dockerfile
dlpierce: Yes, you are correct this is recreating the build process. Not ideal to do ourselves, but I could not find any readily available updated bookworm 2.7.8 images.

Dockerfile
dlpierce: Node 20 appears to already be out of support, and 24 seemed to work for me locally.

cjcolvar (approved):
Looks good!

Dockerfile
cjcolvar: I made a PR for bumping this in circleci. https://github.com/IU-Libraries-Joint-Development/essi/pull/746

cjcolvar:
I'm in favor of going ahead and merging this and following up in another PR to try and fix the flaky tests.

PR #746, opened 2026-07-08
Test with nodejs 24

cjcolvar:
#742 only bumped nodejs to 20.x which is EOL.  This PR brings it up to 24.x and in line with the docker image being built in #745

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,7 @@ jobs:
       - checkout
       - node/install:
           install-yarn: true
-          node-version: '20.20'
+          node-version: '24'
       - browser-tools/install-chrome
       - browser-tools/install-chromedriver
 


cjcolvar:
Looks like this worked except for flaky feature test failures.

aploshay:
@cjcolvar this looks like it's still applicable but as there have been related PRs merged in the meantime, do you want this updated with current `main`?  Or, rebased on it?

cjcolvar:
Just rebased it.  We'll see how the tests go now.

cjcolvar:
Still had a feature failure.  FWIW Looks like it is the same failure in a different spec test than the last CI build in main.

aploshay:
Yeah, #741 also failed on a (different) feature spec :(

aploshay:
Not sure why this line would be failing intermittently:
`./spec/models/bulkrax/mets_xml_entry_spec.rb:100`
My best guess would be the allinson_flex profile not loading, but I'm not sure why that behavior would be inconsistent.

cjcolvar:
I looked at that test a little but decided I didn't know enough to try and remedy it.  It seems to fail less frequently than the feature specs so maybe it isn't worth fighting now.  I restarted the build and it passed this time.

aploshay (approved):

PR #748, opened 2026-07-10
ESSI-2209 OCR and search within issues

randalldfloyd:


aploshay:
@randalldfloyd I think dropping this commit might restore desired behavior:
[[ESSI-2209] [ESSI-2261] Drop uv_search_param call from partial? 1fe0086](https://github.com/IU-Libraries-Joint-Development/essi/pull/748/commits/1fe0086c71b09d2406aa5a94372b33bc3e32d168)

aploshay:
@randalldfloyd updates:
* rebased on current `main`
* I dropped the pair of commits preparing to patch, then patching, `#uv_search_param` -- unfortunately, [the way iiif_print is patching their method in](https://github.com/notch8/iiif_print/blob/b804f165ef614a84acdd239851a1ee3e35c39854/lib/iiif_print/engine.rb#L44), our extensions-based patching didn't do anything
* instead: in [‎app/views/hyrax/base/iiif_viewers/_universal_viewer.html.erb‎](https://github.com/IU-Libraries-Joint-Development/essi/pull/748/changes#diff-a1275aa7018efc10753913dfa4a1b76c6cacaeaf875f5293b04f3a3482bc9308), replace the call to `#uv_search_param` with the replacement method logic we were trying to patch in, directly in the partial
* very minor change to the BlacklightIiifSearch patch, adding the "|| {}" on line 48 to catch `nil` cases

aploshay (approved):

PR #749, opened 2026-07-10
bugfix: stabilize flaky feature specs

aploshay:
* uses `shared_examples` for body of create work feature specs
* attempt to stabilize flaky check by first calling a page refresh, and upping wait time to 5 seconds


randalldfloyd (approved):
+1 This seems worth merging just to get all that duplicative scaffolding into shared examples, whether the wait time helps the feature spec or not.

PR #750, opened 2026-07-13
Run `yarn upgrade` to fix build.

dlpierce:
Running `yarn upgrade` resolves the image build error `Command "webpack" not found.`

There is a lot of noise in the diff from switching to yarnpkg.com, so I'm not sure which specific change makes the difference.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,13 +10,13 @@
     "lodash": "^4.18.1",
     "prop-types": "^15.7.2",
     "react": "^16.13.1",
-    "react_ujs": "^2.6.1",
     "react-bootstrap": "^1.0.1",
     "react-dom": "^16.13.1",
     "react-jsonschema-form": "^1.8.1",
     "react-jsonschema-form-bs4": "^1.7.1",
     "react-spinners": "^0.8.1",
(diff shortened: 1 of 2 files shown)


cjcolvar (approved):
That's a lot of JS dependency changes but we probably need to do manual regression testing due to the nodejs upgrade so now seems like an okay time for this yarn upgrade.

dlpierce:
After more investigation I found the issue is not fixed by a package upgrade, but instead caused by a direct reference to a git+ssh source in yarn.lock which would only be updated when doing a full `yarn upgrade`. Specifically:
```
 http-status-codes@edsilv/http-status-codes#v0.0.7:
   version "0.0.7"
-  resolved "git+ssh://git@github.com/edsilv/http-status-codes.git#af3a8ab9970b1f291a79b03c7382dc16aebee749"
+  resolved "https://codeload.github.com/edsilv/http-status-codes/tar.gz/af3a8ab9970b1f291a79b03c7382dc16aebee749"
```

However, I think you are correct that this is a good time to do a general upgrade due to the Node upgrade.

PR #751, opened 2026-07-15
Manually sleep to avoid flaky tests

cjcolvar:
This adds ~80s to the test suite but it seemed like it helped when I manually ran the feature tests in circleci.


randalldfloyd (approved):
+1 Seems reasonable to expect feature tests to wait a bit on enqueued jobs

PR #752, opened 2026-07-16
Set SSL ENV needed for ruby to find SSL certificates

cjcolvar:
Faraday isn't able to make any SSL connections due to ruby not being able to find the system certificates.  Setting these environment variables seemed to resolve it for me locally.  This should fix dcadmin-devel which can't start up since it tries to make an SSL call to solr while it is booting.


randalldfloyd (approved):
+1 Fixed original symptom discovered of not being able to validate CAS tickets against the IDP from a local dev environment.

dlpierce:
Sorry, I think this is fallout from the recent base image change. My understanding is there are certs embedded into ruby that get used by default by some networking libraries like Net::HTTP

PR #755, opened 2026-07-29
Adding a web proxy that can present a Cloudflare Turnstile challenge

randalldfloyd:
#### There are several parts to this work:

1. The addition of our Python FastAPI-based Turnstile proxy, and its dependencies, into the main Docker image
2. The addition of a Docker compose setup that demonstrates the proxy in an _external container in front of the application container_
3. The addition of `supervisord` into the main image to manage both the proxy and Rails process in a single container setup
4. The addition of a Docker compose setup that demonstrates the proxy _within the application container_

#### Testing the the proxy in an external container:
```
git checkout turnstile-proxy
docker compose -f docker-compose-turnstile-proxied.yml up
```
You should get a Turnstile challenge on the [Catalog endpoint](https://localhost:3000/catalog) (searches, facets etc.)  but nowhere else.

#### Testing the the proxy within the application container:
```
git checkout turnstile-proxy
docker compose -f docker-compose-turnstile-internal.yml up
```
You should get a Turnstile challenge on the [Catalog endpoint](https://localhost:3000/catalog) (searches, facets etc.)  but nowhere else.


aploshay:
+1 for current build running successfully locally, both `internal` and `proxied` versions, and `/catalog` calls getting routed through turnstile, as of:
[827ce1e](https://github.com/IU-Libraries-Joint-Development/essi/pull/755/commits/827ce1ecb51c2728f7f9349efc5b6b92a8011a88)

Dockerfile
```diff
@@ -191,7 +197,7 @@ CMD sidekiq
 # webserver image
 FROM essi-deps AS essi-web
 USER essi:essi
-RUN bundle exec rake assets:precompile
+#RUN bundle exec rake assets:precompile
```
dlpierce: Isn't this needed for deployments where there is no local bind mount of the essi directory? Or in production mode?

randalldfloyd: This was a mistake - I was having problems on that step on my local build but then forgot to uncomment.

